Issue #13, opened 2023-10-20
Revise implicit tagging


JesusMcCloud:
* I think my logic was flawed here: https://github.com/a-sit-plus/kmp-crypto/blob/82dccc97817fed4cc1bc79e194bd86942cadf61b/datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/asn1/BERTags.kt#L65
Can't we implicitly tag CONSTRUCTED ASN.1 Objects too? 
* Do we want to have the the [tag override](https://github.com/a-sit-plus/kmp-crypto/tree/feature/bitstring//datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/asn1/Asn1Encodable.kt#L49) for all `Asn1Decodables`? and merge this extended functionality in the base interface? As we have it now, this implicit tagging only makes sense for primitives, but this is obviously related to the first bullet.

Thoughts, anyone?!

JesusMcCloud:
Turns out we can tag CONSTRUCTED objects too, but we need to take care that primitives cannot be tagged such that the CUNSTRUCTED bit is set too. So there need to be separate functions for CONSTRCUTED and primitive ASN.1 elements (and the tag computation works differently)

I imagine creating implicitly tagged objects to work as follows: 
* `someAsn1Element.tagImplicitly(7u)` in contrast to explicitly tagging (`Asn1.Tagged(7) {+someAsn1Element}`)

Analogously, I imagine decoding implicitly tagged elements as follows:
* `someAsn1Element.verifyImplicitTag<Asn1Sequence>(7)` -- as inline function with reified generic types.

In theory `someAsn1Element.stripImplicitTag<Asn1Primitive>()` is a thing, but where would that make sense?.

Please, @iaik-jheher and @nodh,  let me know what you think about this

JesusMcCloud:
#117 solves many issues. while not too efficient in creating implicitly tagged elements after they've been constructed, `someAsn1Element.tagImplicitly()` has the advantage, that a single function on the base `Asn1Structure` and `Asn1Primitive` suffices to achieve all goals and would let you even override the constructed bit, if desired. This would then return either an Asn1Primitive or an Asn1CustomStructure with the same content as the original element.

`verifyImplicitTag(forceParseToStructure=false)` would also work out nicely. and for the base case where the constructed bit has not been overridden, no magic whatsoever is required.

nodh:
> Please, @iaik-jheher and @nodh, let me know what you think about this

[...]

PR #14, opened 2023-10-20
Feature/bitstring

JesusMcCloud:
Real BIT STRING support, based on KmmBitSet  (which was also reworked).  Required augmenting [Asn1Encodable] (see also #13)


JesusMcCloud:
moved to larger PR (with merge-commit)

PR #24, opened 2023-10-29
Fix and beautify OCTET STRING class hierarchy

JesusMcCloud:
enables processing as such:

```kotlin
val TbsCertificationRequest.attestationProof: List<ByteArray>?
    get() = runCatching {
        attributes?.find { it.oid == OID_ATTESTATION_PROOF }?.value?.singleOrNull()
            .let { it as Asn1Sequence }.children.map { (it as Asn1OctetString<*>).content }
    }.getOrNull()
```


nodh (approved):
nice!

JesusMcCloud:
it's already in `development`

PR #26, opened 2023-11-07
Feature/kmmresult

n0900:
Adds error handling using KmmResult in the following way.
- If conversion from IO or serialization then throw errors (Multibase helper funcitons are assumed to deal with IO or serialization results and as a result throw as well)
- If conversation from one data class to another (i.e. CryptoPublicKey to CoseKey or vice versa) then return KmmResult

Contains some restructuring for clarity.
Test cases have been updated.


CHANGELOG.md
```diff
@@ -16,3 +16,15 @@
  * Initial pretty-printing of ASN.1 Strucutres
  * Massive ASN.1 builder DSL streamlining
  * More convenient explicit tagging
+
+
+### NEXT
+* COSE Support
+* Full RSA and HMAC Support
+* new interface `Asn1OctetString` to unify both ASN.1 OCTET STREAM classes
```
nodh: I would keep it consistent and start with an uppercase letter

.gitmodules
nodh, line 1: Why is this an empty file and checked in?

build.gradle.kts
```diff
@@ -1,7 +1,7 @@
 import org.jetbrains.dokka.gradle.DokkaMultiModuleTask
 
 plugins {
-    id("at.asitplus.gradle.conventions") version "1.9.10+20230922" //version can be omitted for composite builds
+    id("at.asitplus.gradle.conventions") version "1.9.10+20231030" //version can be omitted for composite builds
```
nodh: The commits contained something about Kotlin 1.9.20 ... but here it is still 1.9.10?

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/io/Encoding.kt
```diff
@@ -152,14 +181,13 @@ object MultibaseHelper {
 
     @Deprecated("Dependency of calcEncPublicKeyCoords - Use [multiKeyGetKty] instead ")
```
nodh, line 182: if this method is private, why not migrate it to the mentioned other method?

nodh (changes requested):

nodh:
I've started the review and then someone did change the target branch ...

n0900:
> I've started the review and then someone did change the target branch ...

My bad, correct request now

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/io/Encoding.kt
n0900: Fixed

.gitmodules
n0900: Fixed

CHANGELOG.md
n0900: Fixed and expanded

JesusMcCloud:
Polished some minor rough edges.
The KmmResult-encapsulated `deserialize` functions in the compantions of `CoseKey` and `JsonWebKey` seems off, but they actually make a lot of sense in the context of the VCLib, as you can't run into problems with uncaught exceptions on iOS.

@nodh maybe add such methods to `Asn1Encodable` and `Asn1Decodable` as the third and final option too? `derDecodeSafe()`, maybe?!

nodh:
> Polished some minor rough edges. The KmmResult-encapsulated `deserialize` functions in the compantions of `CoseKey` and `JsonWebKey` seems off, but they actually make a lot of sense in the context of the VCLib, as you can't run into problems with uncaught exceptions on iOS.
> 
> @nodh maybe add such methods to `Asn1Encodable` and `Asn1Decodable` as the third and final option too? `derDecodeSafe()`, maybe?!

Sounds good!

nodh (approved):

PR #29, opened 2023-11-09
Added Symmetric Key Functionality

n0900:
Extended COSE/JOSE Data structure to now handle symmetric key as input


JesusMcCloud:
You mentioned JWKs also support symmetric keys? why not add them as well?

n0900:
> You mentioned JWKs also support symmetric keys? why not add them as well?

adding a 'k'-value should be enough to provide this, no?

JesusMcCloud:
> > You mentioned JWKs also support symmetric keys? why not add them as well?
> 
> adding a 'k'-value should be enough to provide this, no?

You tell me… I'm not the one who read the RFC ;-)

JesusMcCloud (changes requested):
Please do the same for JsonWebKey

JesusMcCloud (approved):

PR #37, opened 2023-12-14
Release/2.2.0

JesusMcCloud:
* Dependency Updates
  * KmmResult 1.5.4
* Refactor `MultiBaseHelper` to only handle conversion
* Change `JwsHeader.publicKey` from JsonWebKey to CryptoPublicKey
* Remove `SignatureValueLength` parameters from JWS- & Cose Algorithm Enum class
* Remove deprecated functions
* Rename `Jws` classes
  * New `CryptoAlgorithm` class 
  * New `CryptoSignature` class for easy Asn1 - RawByteArray conversion
* Rename function in file `JcaExtensions.kt` from `.toPublicKey` to `.toJcaPublicKey` to reflect connection to JVM


datatypes-cose/build.gradle.kts
```diff
@@ -9,7 +8,7 @@ plugins {
     id("at.asitplus.gradle.conventions")
 }
 
-version = "2.1.0"
+version = "2.2.0-SNAPSHOT"
```
nodh, line 11: So its still a snapshot or a release yet?

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsSigned.kt
```diff
@@ -29,31 +33,38 @@ data class JwsSigned(
...
+                } ?: return null.also { Napier.w("Could not parse JWS: $it") }
+
+            return JwsSigned(header, payload, signature)
         }
     }
-}
\ No newline at end of file
+}
+
+fun prepareJwsSignatureInput(header: JwsHeader, payload: ByteArray): String =
```
nodh, line 69: If `header` now gets serialized differently than it was parsed, the signature input is different, and the signature will not match ...

datatypes/build.gradle.kts
```diff
@@ -9,7 +8,7 @@ plugins {
     id("at.asitplus.gradle.conventions")
 }
 
-version = "2.1.0"
+version = "2.2.0"
```
nodh, line 11: So here it is not a snapshot

datatypes/src/jvmMain/kotlin/at/asitplus/crypto/datatypes/JcaExtensions.kt
```diff
@@ -19,21 +19,21 @@ import java.security.interfaces.RSAPublicKey
...
+        CryptoAlgorithm.ES384 -> "SHA384withECDSA"
+        CryptoAlgorithm.ES512 -> "SHA512withECDSA"
+        CryptoAlgorithm.HS256 -> "HmacSHA256"
+        CryptoAlgorithm.HS384 -> "HmacSHA384"
+        CryptoAlgorithm.HS512 -> "HmacSHA512"
+        CryptoAlgorithm.RS256 -> "SHA256withRSA"
+        CryptoAlgorithm.RS384 -> "SHA384withRSA"
+        CryptoAlgorithm.RS512 -> "SHA512withRSA"
+        CryptoAlgorithm.PS256 -> "SHA256withRSA"
+        CryptoAlgorithm.PS384 -> "SHA384withRSA"
```
nodh, line 34: Shouldn't `PS256` be `SHA256withRSAPSS` (and so on)?

nodh (changes requested):

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsSigned.kt
JesusMcCloud: You need to help me understand…

datatypes/src/jvmMain/kotlin/at/asitplus/crypto/datatypes/JcaExtensions.kt
JesusMcCloud: is it called that? I know BC has `…withRSA/PSS`, but is that standardised?

nodh: It seems to be called simply `RSASSA-PSS`, see https://docs.oracle.com/en/java/javase/11/docs/specs/security/standard-names.html and https://github.com/jwtk/jjwt/pull/554

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsSigned.kt
nodh: The JWS compact serialization (the string) gets deserialized into an object of `JwsSigned`, and then its signature gets verified (e.g. in https://github.com/a-sit-plus/kmm-vc-library/blob/main/vclib/src/commonMain/kotlin/at/asitplus/wallet/lib/jws/JwsService.kt#L219). This signature verification should use the string values as they've been serialized (i.e. in the string input), so that the signature will match. But now this original input string is missing from `JwsSigned`, so I'll guess, the signature can't be verified everytime (e.g. when `header.serialize()` now yields a slightly different representation)

JesusMcCloud:
was premature

PR #38, opened 2023-12-14
Fix JWS parsing

nodh:
Always take the unmodified serialization as input for signature
verification, there is no canoicalization in JWS


datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsSigned.kt
```diff
@@ -14,13 +14,10 @@ import io.matthewnelson.encoding.core.Encoder.Companion.encodeToString
 data class JwsSigned(
     val header: JwsHeader,
     val payload: ByteArray,
-    val signature: CryptoSignature
+    val signature: CryptoSignature,
+    val plainSignatureInput: String,
```
JesusMcCloud, line 18: do we want to keep this private?

nodh: No, because otherwise the `JwsService` from vclib can't access it

JesusMcCloud (approved):

PR #48, opened 2024-02-29
Fix cose key serialization

JesusMcCloud:
finally implement fully working EC-point-compression-agnostic COSE key serialization


JesusMcCloud:
Once this is approved, I'll push out a new release and bump kmp-crypto in vclib

nodh (approved):
Well, if it really works now, fine!

PR #55, opened 2024-04-08
Update JWS, JWT implementation

nodh:
Update data classes as a prerequisite to https://github.com/a-sit-plus/kmm-vc-library/pull/48

Also cc'ing @n0900

### DIFF
--- a/datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JsonWebKey.kt
+++ b/datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JsonWebKey.kt
@@ -27,14 +27,12 @@ data class JsonWebKey(
     val type: JwkType? = null,
     @SerialName("kid")
     val keyId: String? = null,
-    //EC
     @SerialName("x")
     @Serializable(with = ByteArrayBase64UrlSerializer::class)
     val x: ByteArray? = null,
     @SerialName("y")
     @Serializable(with = ByteArrayBase64UrlSerializer::class)
     val y: ByteArray? = null,
-    //RSA
     @SerialName("n")
     @Serializable(with = ByteArrayBase64UrlSerializer::class)
     val n: ByteArray? = null,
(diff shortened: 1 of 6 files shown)


datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JsonWebKey.kt
```diff
@@ -27,14 +27,12 @@ data class JsonWebKey(
     val type: JwkType? = null,
     @SerialName("kid")
     val keyId: String? = null,
-    //EC
```
JesusMcCloud, line 30: Why remove the comment?

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsHeader.kt
```diff
@@ -45,20 +45,50 @@ data class JwsHeader(
...
+                ?.firstNotNullOfOrNull {
+                    runCatching {
+                        X509Certificate.decodeFromTlv(Asn1Element.parse(it) as Asn1Sequence).publicKey
+                    }.getOrNull() ?: runCatching {
+                        X509Certificate.decodeFromTlv(Asn1Element.parse(it.decodeX5c()) as Asn1Sequence).publicKey
+                    }.getOrNull()
+                }
+    }
+
+    private fun ByteArray.decodeX5c() = decodeToString()
```
JesusMcCloud, line 78: A function like that should not be just a private utility in KMP-Crypto.
It would make more sense to add it to the certificate class, for example, because it provides added value in general.

JesusMcCloud (changes requested):

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JsonWebKey.kt
nodh: Because it's not a KDoc comment, so it's not visible from outside the source code, and it provides no added value.

JesusMcCloud: since this class mixes up rsa and ec properties as per the spec, I do feel it provides added value to maintainers of the code base

JesusMcCloud (approved):

PR #56, opened 2024-04-17
Add CBOR Web Token

nodh:
Implement CBOR Web Token from [RFC 8392](https://datatracker.ietf.org/doc/html/rfc8392), to be used in [OpenID for Verifiable Credenital Issuance](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#authorization-details)


datatypes-cose/src/commonMain/kotlin/at/asitplus/crypto/datatypes/cose/InstantLongSerializer.kt
JesusMcCloud, line 1: This one looks familiar. Is is really nowhere to be found?

JesusMcCloud (approved):
Please just double-check that the serializer cannot be pulled from somewhere else

PR #63, opened 2024-05-13
Add extensions for converting from/to bouncycastle

iaik-jheher:


iaik-jheher:
TODO: also for `CryptoSignature`

JesusMcCloud (changes requested):
Code looks good!
Set development branch as target and amend the changelog, please

iaik-jheher:
after additional discussion, we want to do this differently as part of a restructure of the EC classes

PR #65, opened 2024-05-15
Rename EcCurve to ECCurve and CryptoPublicKey.Ec to CryptoPublicKey.EC

iaik-jheher:
as per alignment with @JesusMcCloud, this brings them in line with `CryptoSignature.EC`


JesusMcCloud (approved):
lgtm

PR #66, opened 2024-05-15
Implement ASN.1 encoding/decoding for BigInteger

iaik-jheher:


JesusMcCloud (changes requested):
missing changelog, otherwise fine

JesusMcCloud (approved):
ship it!

PR #67, opened 2024-05-15
add missing parameters for EC curves

iaik-jheher:


datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/ECCurve.kt
```diff
@@ -15,87 +16,162 @@ import kotlinx.serialization.descriptors.SerialDescriptor
...
+
+    /**
+     * h: Cofactor of the cyclic subgroup generated by G
+     * See https://www.secg.org/sec2-v2.pdf
+     */
+    val cofactor: Int get() =
+        when(this) {
+            SECP_256_R_1 -> 1
+            SECP_384_R_1 -> 1
+            SECP_521_R_1 -> 1
```
JesusMcCloud, line 172: those can be condensed into a single comma-separated case

JesusMcCloud (approved):
As always, very thorough.

Condense the `when` at your discretion

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/ECCurve.kt
iaik-jheher: stylistically i'd prefer to keep it separate

PR #69, opened 2024-05-22
Typo fix: deprecated replacement getter

iaik-jheher:
cf. #67


iaik-jheher:
is there a reason test cases don't automatically run on PRs? this could've/should've been caught before merging

JesusMcCloud:
> is there a reason test cases don't automatically run on PRs? this could've/should've been caught before merging

because I did not set them up to run. they only run on push

JesusMcCloud (approved):

PR #71, opened 2024-05-23
CryptoSignature rework

iaik-jheher:
* Store `CryptoSignature` as a pair of integers, instead of byte arrays.
* Disallow raw encoding unless the underlying curve's scalar byte length is known.
* Require explicit opt-in (`guessCurve`) before guessing byte length of the underlying curve.
  * Or allow the curve to be specified from context (`withScalarByteLength`, `withCurve`)


JesusMcCloud (approved):
lgtm

PR #72, opened 2024-05-23
Unit test our dependencies

iaik-jheher:
...because they clearly aren't doing it themselves.


iaik-jheher:
cf ionspin/kotlin-multiplatform-bignum#291 and ionspin/kotlin-multiplatform-bignum#292

JesusMcCloud:
ship it

PR #79, opened 2024-06-03
Fix x.509 EC length

iaik-jheher:
We currently assume that X.509 certificates' signatures use the "native" curve for the hash algorithm. For example, we assume that signatures using `ecdsa-with-SHA256` will be on p256.

There is no such requirement in the spec, and many certificates in practice do not do this, and a sanity check for bit length in the definitive length signature constructor exposes this in a variety of test cases using real-world certificates.

X.509 certificates carry no information about their signature's underlying curve. (Because X.509 signatures also do not.)
This information is derived from the public key used for verification, but this is not available when parsing a certificate. (You would need the next certificate in the chain.)

Thus, X.509 EC signatures are not `IndefiniteLength`. The only thing this means is that they cannot be converted to raw bytes without providing contextual information (i.e., the underlying curve).


iaik-jheher:
Also changed what "equality" means for `CryptoSignature` instances; different encodings of the same signature now compare equally.

iaik-jheher:
CI seems to be broken in mysterious ways. Pretty sure that wasn't me.

CHANGELOG.md
JesusMcCloud, line 1: That's not how a changelog works

JesusMcCloud (changes requested):
Fix the changelog to be a changelog, please

PR #80, opened 2024-06-03
Clean up compressed-representation-preferred flag on `CryptoPublicKey`

iaik-jheher:
Right now, we have `var useCompressedRepresentation`, which does two things:

- Store whether the key was originally decoded from compressed, to preserve from-to binary compatibility for COSE
- Signal to `toAnsiX963Encoded` whether it should return compressed representation

This means that the only valid way to obtain a particular form (encoded/unencoded) on demand is to store the var's value, replace it, call the function, then put the var back. This makes no sense.


datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/CryptoPublicKey.kt
```diff
@@ -321,9 +321,9 @@ sealed class CryptoPublicKey : Asn1Encodable<Asn1Sequence>, Identifiable {
          */
         override val didEncoded by lazy {
             val codec = when (curve) {
-                ECCurve.SECP_256_R_1 -> if (useCompressedRepresentation) 0x1200u else 0x1290u
-                ECCurve.SECP_384_R_1 -> if (useCompressedRepresentation) 0x1201u else 0x1291u
-                ECCurve.SECP_521_R_1 -> if (useCompressedRepresentation) 0x1202u else 0x1292u
+                ECCurve.SECP_256_R_1 -> if (preferCompressedRepresentation) 0x1200u else 0x1290u
```
iaik-jheher, line 324: @nodh Bernd asks whether these fake codecs (`0x129_`) are actually needed for anything, or why we shouldn't always canonically serialize DIDs in compressed form which would give us interop with everything else.

nodh: If we now have compressed representation on all platforms, we should switch to the canonical prefix of `0x1200u`, yes!

JesusMcCloud (approved):

PR #82, opened 2024-06-10
Release/3.2.0

JesusMcCloud:
* Kotlin 2.0
* Gradle 8.8
* Bouncy Castle 1.78.1
* Kotest 5.9.1 
* Coroutines 1.8.1
* Serialization 1.7.1-SNAPSHOT

**Fixes**
 * Move `curve` from `CryptoAlgorithm` to `JwsAlgorithm`
 * Don't read curve information into the X.509 tea leaves where none exists
   * `CryptoSignature`s in X.509 are now indefinite length
 * Rename `CryptoAlgorithm` to `X509SignatureAlgorithm` to better describe what it is


datatypes-cose/src/commonMain/kotlin/at/asitplus/crypto/datatypes/cose/CoseAlgorithm.kt
```diff
@@ -37,23 +37,23 @@ enum class CoseAlgorithm(val value: Int) {
 
 
     fun toCryptoAlgorithm() = when(this) {
```
nodh, line 39: But now the function name does not match the class name of the return values

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsAlgorithm.kt
```diff
@@ -41,23 +39,32 @@ enum class JwsAlgorithm(override val identifier: String):JsonWebAlgorithm {
     NON_JWS_SHA1_WITH_RSA("RS1");
 
     fun toCryptoAlgorithm() = when (this) {
```
nodh, line 41: Same here with the name

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/CryptoPublicKey.kt
```diff
@@ -71,11 +71,11 @@ sealed class CryptoPublicKey : Asn1Encodable<Asn1Sequence>, Identifiable {
             val keyBytes = decoded.copyOfRange(2, decoded.size)
 
             return when (codec) {
-                0x1200uL, 0x1290uL ->
```
nodh, line 74: Why don't we still support our inofficial values?

datatypes/src/iosTest/kotlin/Test.kt
```diff
@@ -0,0 +1,9 @@
+import io.kotest.core.spec.style.FreeSpec
+import io.kotest.matchers.shouldNotBe
+
+class Test: FreeSpec( {
+
+    "This dummy test" {
+       "is just making shure" shouldNotBe "that iOS tests are indeed running"
```
nodh, line 7: I see, you did even include a typo to make sure it's really your test that runs!

nodh (changes requested):

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/CryptoPublicKey.kt
JesusMcCloud: I was under the impression this was sorted out with @iaik-jheher

datatypes/src/iosTest/kotlin/Test.kt
JesusMcCloud: https://www.shure.com/

CHANGELOG.md
```diff
@@ -138,3 +138,19 @@
...
+* Bouncy Castle 1.78.1
+* Kotest 5.9.1 
+* Coroutines 1.8.1
+* Serialization 1.7.1-SNAPSHOT
+
+**Fixes**
+ * Move `curve` from `CryptoAlgorithm` to `JwsAlgorithm`
+ * Don't read curve information into the X.509 tea leaves where none exists
+   * `CryptoSignature`s in X.509 are now indefinite length
+ * Rename `CryptoAlgorithm` to `X509SignatureAlgorithm` to better describe what it is
```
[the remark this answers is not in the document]
JesusMcCloud: because what's contained in there conforms to X509 (in contrast to JWS oder COSE).
The key "missing" information for EC signatures, for example, is the curve and our encoding/decoding function for what used to be called. `CryptoSignature` were strictly following X509, so @iaik-jheher convinced me that a rename was in order. Also: we are only dealing with signatures in this class. So if we expand this to encryption, it makes perfect sense to separate this (like we do with JWA already)

datatypes-cose/src/commonMain/kotlin/at/asitplus/crypto/datatypes/cose/CoseAlgorithm.kt
JesusMcCloud: renamed

datatypes-jws/src/commonMain/kotlin/at/asitplus/crypto/datatypes/jws/JwsAlgorithm.kt
JesusMcCloud: renamed

JesusMcCloud:
Yeah, the API needs to be revised, because of EC signatures now being handled correctly (i.e. without guessing).

datatypes/src/commonMain/kotlin/at/asitplus/crypto/datatypes/CryptoPublicKey.kt
nodh: Encoding just the "official" values is fine, but why do we need to drop support for decoding it?

PR #86, opened 2024-06-12
Provider: Signature verification

iaik-jheher:


provider/src/commonMain/kotlin/at/asitplus/crypto/provider/sign/SignatureInput.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    constructor(data: ByteArray) : this(sequenceOf(data), RAW_BYTES)
+    constructor(data: Sequence<ByteArray>): this(data, RAW_BYTES)
+
+    /**
+     * Takes the leftmost bits of the byte array, and converts them to an unsigned `BigInteger`.
+     *
+     * (This matches the ECDSA spec.)
+     */
+    fun asBigInteger(length: BitLength): BigInteger {
```
JesusMcCloud, line 43: KmmResult?

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/sign/Verifier.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    sealed class EC
+    @Throws(IllegalArgumentException::class)
+    constructor (
+        final override val signatureAlgorithm: SignatureAlgorithm.ECDSA,
+        final override val publicKey: CryptoPublicKey.EC)
+    : Verifier {
+        init {
+            signatureAlgorithm.requiredCurve?.let {
+                require(publicKey.curve == it)
```
JesusMcCloud, line 26: Can we McGuyer a creation function that returns a KmmResult? not sure myself whether we want this

JesusMcCloud (commented):
looking good so far

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/sign/Verifier.kt
iaik-jheher: i don't think this is a good idea in the constructor, but i also don't think the constructor will be the "canonical" way to construct verifiers. once we end up with `X509Algorithm.EC.verifierFor(publicKey)` then that can return `KmmResult`

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/sign/SignatureInput.kt
iaik-jheher: there's no useful error conditions to catch

JesusMcCloud: Soooo it's an IO error?!

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/sign/Verifier.kt
JesusMcCloud: Then be sure to add API Docs, referencing `X509Algorithm.EC.verifierFor(publicKey)` once its there.

iaik-jheher: Sure, the constructor will likely end up internal anyway.

PR #91, opened 2024-06-26
Fix/cose signature

nodh:
Quick fix for Release 3.2.0 to correctly serialize COSE sign structure.


nodh:
I'll suggest to create a release 3.2.1 with the fix

JesusMcCloud (approved):

PR #94, opened 2024-07-08
Provider: Signing

iaik-jheher:


provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
```diff
@@ -247,33 +265,61 @@ object IOSKeychainProvider:  TPMSigningProviderI<iosSigner<*>, iosSignerConfigur
...
+                            kSecAttrAccessible mapsTo availability
+                        }
+                        else -> {
+                            kSecAttrAccessControl mapsTo corecall {
+                                SecAccessControlCreateWithFlags(
+                                    null, availability,
+                                    when {
+                                        (factors.biometry && factors.deviceLock) -> kSecAccessControlUserPresence
+                                        factors.biometry -> kSecAccessControlBiometryAny
+                                        else -> kSecAccessControlDevicePasscode
```
JesusMcCloud, line 304: what if i want other flags? what if the assumptions here don't hold for my use case?

JesusMcCloud: anyways, nested cfdictionaries are segfaulting (no clue why). When constructing them in swift, the segfault goes away. This was not triggered on the simulator, since the combination of flags caused it to not get accesses through key generation (beacuse kSecAccessControl was never evaluated)

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
```diff
@@ -186,9 +193,13 @@ sealed class iosSigner<H : unlockedIOSSigner>(
                         is ECDSA -> kSecAttrKeyTypeEC
                         is RSA -> kSecAttrKeyTypeRSA
                     },
+                    kSecMatchLimit to kSecMatchLimitOne,
                     kSecReturnRef to true,
 
-                    kSecUseOperationPrompt to config.unlockPrompt.v.message,
+                    kSecUseAuthenticationContext to LAContext().apply {
```
JesusMcCloud, line 207: we will need to go down two routes here: on the on hand keeping the context and emulating Android-Style timeout as in my original swift-based approach, so that key with biometric unlock works the same on both platform with and without timeout. on on the other hand, allow for low-level customisation as in my original swift approach, so all bases are covered.

as for attestation: In the generic case, fix it to P256 and P384 so both platforms behave the same and on android,  i think we can already freely customise, because this is anyway a superset of what we can emulate on iOS

provider/src/androidMain/kotlin/at/asitplus/crypto/provider/os/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,402 @@
...
+import kotlin.enums.enumEntries
+import java.security.Signature as JCASignatureObject
+
+internal sealed interface FragmentContext {
+    @JvmInline value class OfActivity(val activity: FragmentActivity): FragmentContext
+    @JvmInline value class OfFragment(val fragment: Fragment): FragmentContext
+}
+
+
+class AndroidTPMConfiguration internal constructor(): PlatformSigningKeyConfiguration.TPMConfiguration() {
```
JesusMcCloud, line 57: `TMP` -> `SE`. this applies to the whole codebase

provider/src/androidMain/kotlin/at/asitplus/crypto/provider/os/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,402 @@
...
+    /** Whether a StrongBox TPM is required. */
+    var requireStrongBox = false
+}
+class AndroidSigningKeyConfiguration internal constructor(): PlatformSigningKeyConfiguration<AndroidSignerConfiguration>() {
+    override val tpm = childOrNull(::AndroidTPMConfiguration)
```
JesusMcCloud, line 62: se

provider/src/androidMain/kotlin/at/asitplus/crypto/provider/os/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,402 @@
...
+                throw IllegalArgumentException("Key supports multiple ${what}s ($valid). You need to specify $what in signer configuration.")
+            val only = valid.first()
+            possible.find {
+                nameMap(it).equals(only, ignoreCase=true)
+            } ?: throw UnsupportedCryptoException("Unsupported $what $only")
+        }
+    }
+
+sealed class AndroidKeyStoreProviderImpl<SignerT: AndroidKeystoreSigner> private constructor() :
+    TPMSigningProviderI<SignerT, AndroidSignerConfiguration, AndroidSigningKeyConfiguration>
```
JesusMcCloud, line 102: TMP -> SE

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/os/TPMSigningProvider.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    open class ProtectionConfiguration internal constructor(): DSL.Data() {
+        /** The timeout before this key will need to be unlocked again. */
+        var timeout: Duration = 0.seconds
+        /** Which authentication factors can authorize this key;
+         * if multiple factors are specified, any one of them can authorize the key */
+        val factors = child(::ProtectionFactorConfiguration)
+    }
+
+    open class TPMConfiguration: DSL.Data() {
```
JesusMcCloud, line 67: TPM -> SE

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/os/TPMSigningProvider.kt
```diff
@@ -0,0 +1,130 @@
...
+        open val attestation = childOrNull(::AttestationConfiguration)
+        open val protection = childOrNull(::ProtectionConfiguration)
+    }
+
+    open val tpm = childOrNull(::TPMConfiguration)
```
JesusMcCloud, line 72: SE

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
```diff
@@ -0,0 +1,416 @@
...
+    private val tags by lazy {
+        val bundleId = NSBundle.mainBundle.bundleIdentifier
+            ?: throw UnsupportedCryptoException("Keychain access is unsupported outside of a Bundle")
+        Pair("kmp-crypto-privatekey-$bundleId", "kmp-crypto.publickey-$bundleId")
+    }
+    val PRIVATE_KEYS get() = tags.first
+    val PUBLIC_KEYS get() = tags.second
+}
+
+class iosTPMConfiguration internal constructor() : PlatformSigningKeyConfiguration.TPMConfiguration() {
```
JesusMcCloud, line 113: here, tpm makes even less sense, because only parts are secure-enclace protected. but SE (in short for Secure Element) is at least somewhat sensible, because as soon as biometrics are involved, an SE in back in the loop

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
```diff
@@ -0,0 +1,416 @@
...
+class iosSigningKeyConfiguration internal constructor(): PlatformSigningKeyConfiguration<iosSignerConfiguration>() {
+    override val tpm = childOrDefault(::iosTPMConfiguration) {
+        requireSecureEnclave = false
+    }
+}
+
+class iosSignerConfiguration internal constructor(): SignerConfiguration() {
+}
+
+sealed class unlockedIOSSigner(private val ownedArena: Arena, private val privateKeyRef: SecKeyRef) : Signer.UnlockedHandle {
```
JesusMcCloud, line 131: IOS -> Ios for nicer CamelCase

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
JesusMcCloud: Also: While the Secure enclave does not support RSA, the keychain can protect access to those keys using biometrics and passcode to behave the same as enclave-protected key material, so generating RSA keys and protecting them using passcode/biometrics should work. too.

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
iaik-jheher: resolved by 2941621

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
iaik-jheher: This is already possible by explicitly opting out of SE protection (`tpm { requireSecureEnclave = false }`).

I am reluctant to transparently "downgrade" the expected security without explicit opt-in.

provider/src/androidMain/kotlin/at/asitplus/crypto/provider/os/AndroidKeyStoreProvider.kt
iaik-jheher: Is "secure enclave" a generic term that android devs would recognize? I primarily associate it with Apple-world.

Can we maybe find a more generic term... `hardware-backed` or something?

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
iaik-jheher: > keeping the context and emulating Android-Style timeout as in my original swift-based approach, so that key with biometric unlock works the same on both platform with and without timeout

this is something i am very conflicted on... on the one hand, it's definitely _needed_; on the other hand, we are pretending there's security where there's none... a malicious app could reuse our key an arbitrary number of times with a single biometry unlock, even if we set a small/zero reuse timeout in our own code. this is purely theater, the security isn't being enforced anywhere.

demoapp/DEVELOPMENT.md
```diff
@@ -0,0 +1,6 @@
+**REQUIRES a MacOS Host to build all modules**
+
+* recursively clone this repo
+* set `sdk.dir=/absulute/path/to/Android/sdk` inside `kmp-crypto/local.properties`
+* `cd kmp-crypto && ./gradlew publishAllPublicationsToLocalRepository`
```
iaik-jheher, line 5: @JesusMcCloud i don't think this is accurate anymore? i don't remember having to recursively clone anything, or needing to run gradle commands :9

JesusMcCloud: swift-klib is still pulled-in recursively and the swift code is used by the verifier. but publishing is not needed anymore

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IOSKeychainProvider.kt
JesusMcCloud: > This is already possible by explicitly opting out of SE protection (`tpm { requireSecureEnclave = false }`).
> 
> I am reluctant to transparently "downgrade" the expected security without explicit opt-in.

D'Accord!

JesusMcCloud: > > keeping the context and emulating Android-Style timeout as in my original swift-based approach, so that key with biometric unlock works the same on both platform with and without timeout
> 
> this is something i am very conflicted on... on the one hand, it's definitely _needed_; on the other hand, we are pretending there's security where there's none... a malicious app could reuse our key an arbitrary number of times with a single biometry unlock, even if we set a small/zero reuse timeout in our own code. this is purely theater, the security isn't being enforced anywhere.

I partially agree. The weird thing is, you can actually get that timeout behaviour enforced by the SE, but only if you recently **unlocked** your device. Completely bonkers on Apple's side. Moreover? What is your scenario for a malicious app? It will mos def not come from the app store and if you jailbreak your device all bets are off anyways. 

In addition, you can and want to add another layer of theater through attestation and this is as good as it gets on iOS.

provider/src/androidMain/kotlin/at/asitplus/crypto/provider/os/AndroidKeyStoreProvider.kt
JesusMcCloud: SE as in "secue element" hardware-backed is fine too. The code should read nicely in the end

iaik-jheher: renames are done, check when you have time

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/os/SigningProvider.kt
```diff
@@ -0,0 +1,141 @@
...
+    open class ProtectionConfiguration internal constructor(): DSL.Data() {
+        /** The timeout before this key will need to be unlocked again. */
+        var timeout: Duration = 0.seconds
+        /** Which authentication factors can authorize this key;
+         * if multiple factors are specified, any one of them can authorize the key */
+        val factors = childOrDefault(::ProtectionFactorConfiguration)
+    }
+
+    open class SecureHardwareConfiguration: DSL.Data() {
+        /** Whether to use hardware-backed storage, such as Android TPM or Apple's Secure Enclave.
```
JesusMcCloud, line 71: Still says TPM. the rest looks good wrt. renaming

provider/src/iosMain/kotlin/at/asitplus/crypto/provider/os/IosKeychainProvider.kt
```diff
@@ -0,0 +1,420 @@
...
+                            kSecAttrAccessible mapsTo availability
+                        }
+                        else -> {
+                            kSecAttrAccessControl mapsTo corecall {
+                                SecAccessControlCreateWithFlags(
+                                    null, availability,
+                                    when {
+                                        (factors.biometry && factors.deviceLock) -> kSecAccessControlUserPresence
+                                        factors.biometry -> kSecAccessControlBiometryAny
+                                        else -> kSecAccessControlDevicePasscode
```
JesusMcCloud, line 325: I'm pretty sure we want to be able to fully, manually specify this ba accessing those constants and not through a wrapper. but do present your arguments and I may be convinced otherwise

provider/src/commonMain/kotlin/at/asitplus/crypto/provider/os/Attestation.kt
```diff
@@ -0,0 +1,63 @@
+package at.asitplus.crypto.provider.os
```
iaik-jheher, line 1: should this be in datatypes somewhere? open to input

JesusMcCloud: probably

iaik-jheher:
TODO on iOS:
- [ ] check 60s key unlock -> 0s key failure -> 60s key unlock
- [ ] store supported digests & paddings in metadata

iaik-jheher:
todo readme

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProvider.kt
```diff
@@ -366,34 +322,49 @@ sealed class LockedAndroidKeystoreSigner private constructor(
...
+    final override suspend fun trySetupUninterruptedSigning(configure: DSLConfigureFn<AndroidSignerSigningConfiguration>) = catching {
+        if (needsAuthentication && !needsAuthenticationForEveryUse) {
+            getJCASignature(DSL.resolve(::AndroidSignerSigningConfiguration, configure))
+        }
+    }
+
+    final override suspend fun sign(
+        data: SignatureInput,
+        configure: DSLConfigureFn<AndroidSignerSigningConfiguration>
+    ): KmmResult<CryptoSignature> = catching {
```
JesusMcCloud, line 349: I maintain that this should return a custom ternary sealed result class:
* successful signature CryptoSignature is part of the result
* auth failed -> maybe with a string explanation)
* error -> contains a throwable

The same goes for iOS. On the JVM the same could happen for some non-standard procedures, so it does make sense that also the JVM returns such a ternary result.

demoapp/DEVELOPMENT.md
JesusMcCloud, line 5: Note to self: fix this

demoapp/README.MD
JesusMcCloud, line 1: Note to self: check and fix

demoapp/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+    alias(libs.plugins.multiplatform).apply(false)
+    alias(libs.plugins.compose).apply(false)
+    alias(libs.plugins.android.application).apply(false)
+    alias(libs.plugins.buildConfig).apply(false)
+ //   id("at.asitplus.gradle.conventions") version "1.9.23+20240319+1"
```
JesusMcCloud, line 6: note to self: remove

demoapp/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+allprojects {
+    repositories {
+        maven(rootProject.projectDir.absolutePath+"/kmp-crypto/repo")
+        maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
+        maven(uri("https://raw.githubusercontent.com/a-sit-plus/kotlinx.serialization/mvn/repo"))
```
JesusMcCloud, line 13: note to self: remove

demoapp/composeApp/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+    }
+}
+
+android {
+    namespace = "at.asitplus.cryptotest"
+    compileSdk = 34
+
+    defaultConfig {
+        minSdk = 33
+        targetSdk = 34
```
JesusMcCloud, line 80: note to self: check requirements

demoapp/composeApp/build.gradle.kts
```diff
@@ -0,0 +1,111 @@
...
+    }
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
+    }
+    buildFeatures {
+        compose = true
+    }
+    composeOptions {
+        kotlinCompilerExtensionVersion = "1.5.4"
```
JesusMcCloud, line 99: note to self: check!

demoapp/composeApp/src/iosMain/kotlin/at/asitplus/cryptotest/App.ios.kt
JesusMcCloud, line 1: note to self: delete

demoapp/composeApp/src/jvmMain/kotlin/at/asitplus/cryptotest/App.jvm.kt
JesusMcCloud, line 1: note to self: delete

demoapp/settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+        google()
+        gradlePluginPortal()
+        mavenCentral()
+
+        maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
+        maven {
+            url =
+                uri("https://raw.githubusercontent.com/a-sit-plus/gradle-conventions-plugin/mvn/repo")
+            name = "aspConventions"
+        }
```
JesusMcCloud, line 15: note to self: remove

indispensable/build.gradle.kts
```diff
@@ -192,12 +192,6 @@ kotlin {
             }
         }
```
JesusMcCloud, line 194: note to self: remove failing testcases

supreme/src/androidInstrumentedTest/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProviderTests.kt
JesusMcCloud, line 1: `.shouldBeSuccess` would be better than `getOrThrow`

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/hazmat/InternalsAccessors.kt
```diff
@@ -0,0 +1,21 @@
...
+import at.asitplus.signum.supreme.sign.EphemeralKeyBase
+import at.asitplus.signum.supreme.sign.Signer
+import java.security.PrivateKey
+
+/** The underlying JCA [PrivateKey] object. */
+@HazardousMaterials
+val EphemeralKey.jcaPrivateKey get() = (this as? EphemeralKeyBase<*>)?.privateKey as? PrivateKey
+
+/** The underlying JCA [PrivateKey] object. Not available for unlocked KeyStore signers; see [jcaSignatureInstance]. */
+@HazardousMaterials
```
JesusMcCloud, line 16: is this still true and sensible after the rewrite?

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,393 @@
...
+                    PREFERRED -> false // TODO
+                    DISCOURAGED -> false
+                })
+                hw.attestation.v?.let {
+                    setAttestationChallenge(it.challenge)
+                }
+                hw.protection.v?.let {
+                    setUserAuthenticationRequired(true)
+                    setUserAuthenticationParameters(it.timeout.inWholeSeconds.toInt(),
+                        (if (it.factors.v.biometry) KeyProperties.AUTH_BIOMETRIC_STRONG else 0) or
```
JesusMcCloud, line 194: note to self: may need switch depending on android version. this depends on target and minsdk etc.

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProvider.kt
```diff
@@ -0,0 +1,393 @@
...
+    keyInfo.isUserAuthenticationRequired
+val AndroidKeystoreSigner.needsAuthenticationForEveryUse inline get() =
+    keyInfo.isUserAuthenticationRequired &&
+            (keyInfo.userAuthenticationValidityDurationSeconds <= 0)
+
+/*actual typealias PlatformSigningProviderSigner = AndroidKeystoreSigner
+actual typealias PlatformSigningProviderSignerConfiguration = AndroidSignerConfiguration
+actual typealias PlatformSigningProviderSigningKeyConfiguration = AndroidSigningKeyConfiguration
+actual typealias PlatformSigningProvider = AndroidKeyStoreProvider
+actual typealias PlatformSigningProviderConfiguration = PlatformSigningProviderConfigurationBase*/
```
JesusMcCloud, line 391: remove commented-out code

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/Throwables.kt
```diff
@@ -1,6 +1,12 @@
...
+annotation class HazardousMaterials
+
 sealed class CryptoException(message: String? = null, cause: Throwable? = null) : Throwable(message, cause)
-class CryptoOperationFailed(message: String) : CryptoException(message)
+open class CryptoOperationFailed(message: String) : CryptoException(message)
+
+open class UnsupportedCryptoException(message: String? = null, cause: Throwable? = null) : CryptoException(message, cause)
 
-class UnsupportedCryptoException(message: String? = null, cause: Throwable? = null) : CryptoException(message, cause)
+class UnlockFailed(message: String? = null, cause: Throwable? = null) : Throwable(message, cause)
```
JesusMcCloud, line 12: see my other comment. is this an error or auth failed?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/os/SigningProvider.kt
```diff
@@ -0,0 +1,211 @@
...
+        sign(SignatureInput(data), configure)
+    suspend fun sign(data: Sequence<ByteArray>, configure: DSLConfigureFn<SigningConfiguration> = null) =
+        sign(SignatureInput(data), configure)
+    override suspend fun sign(data: SignatureInput) = sign(data, null)
+    override suspend fun sign(data: ByteArray) = sign(SignatureInput(data), null)
+    override suspend fun sign(data: Sequence<ByteArray>) = sign(SignatureInput(data), null)
+}
+
+open class PlatformSigningProviderConfigurationBase internal constructor(): DSL.Data()
+// BLOCKED BY KT-71036
```
JesusMcCloud, line 156: then it is ok to keep the commented-out code

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/sign/Verifier.kt
JesusMcCloud, line 1: I'd change the package name from `sign` to `sig` since signer and verifier live here

supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
```diff
@@ -0,0 +1,628 @@
...
+                        }
+                        else -> {
+                            val factors = hwProtection.factors.v
+                            kSecAttrAccessControl mapsTo corecall {
+                                SecAccessControlCreateWithFlags(
+                                    null, availability,
+                                    when {
+                                        (factors.biometry && factors.deviceLock) -> kSecAccessControlUserPresence
+                                        factors.biometry -> kSecAccessControlBiometryAny
+                                        else -> kSecAccessControlDevicePasscode
```
JesusMcCloud, line 472: I am still not sure about this. on iOS you can chose to require either _any_ or _all_ specified auth methods. But this is probably a case for an issue and not for this PR

JesusMcCloud (changes requested):
Awesome work! I'd really appreciate if some more of the internals would be as well documend as the ObjC-Interop glue. See also my other comments. I'll tend to the ones noted for me and resolve them myself.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/os/SigningProvider.kt
iaik-jheher: #107

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
```diff
@@ -0,0 +1,41 @@
...
+sealed class UserInitiatedCancellationReason(message: String?, cause: Throwable?): Throwable(message, cause)
+class UnlockFailed(message: String? = null, cause: Throwable? = null) : UserInitiatedCancellationReason(message, cause)
+
+sealed interface SignatureResult {
+    @JvmInline
+    value class Success(val signature: CryptoSignature): SignatureResult
+    @JvmInline
+    value class Failure(val problem: UserInitiatedCancellationReason): SignatureResult
+    @JvmInline
+    value class Error(val exception: Throwable): SignatureResult
```
JesusMcCloud, line 18: Add KDoc describing semantics of `Error` and `Failure` to avoid misinterpretation

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
```diff
@@ -0,0 +1,41 @@
...
+}
+val SignatureResult.isSuccess get() = (this is SignatureResult.Success)
+/** Retrieves the contained signature, asserting it exists. If it does not exist, throws the contained problem. */
+val SignatureResult.signature: CryptoSignature get() = when (this) {
```
JesusMcCloud, line 22: Annotate with @throws for good measure

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
```diff
@@ -0,0 +1,41 @@
...
+}
+/** Retrieves the contained signature, if one exists. */
+val SignatureResult.signatureOrNull: CryptoSignature? get() = when (this) {
+    is SignatureResult.Success -> this.signature
+    else -> null
+}
+/** Transforms this SignatureResult into a [KmmResult]. Both [Failure] and [Error] map to [KmmResult.Failure]. */
+fun SignatureResult.wrap(): KmmResult<CryptoSignature> = catching { this.signature }
+
+internal inline fun signCatching(fn: ()->CryptoSignature): SignatureResult =
```
JesusMcCloud, line 35: Even through it is internal. document the which case maps to what using KDoc.

supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
```diff
@@ -328,8 +346,8 @@ sealed class IosSigner(final override val alias: String,
             { "Metadata type mismatch (RSA key, metadata not RSA) "}
 
             signatureAlgorithm = SignatureAlgorithm.RSA(
-                digest = resolveOption("digest", metadata.algSpecific.supportedDigests, config.rsa.v.digestSpecified, config.rsa.v.digest),
-                padding = resolveOption("padding", metadata.algSpecific.supportedPaddings, config.rsa.v.paddingSpecified, config.rsa.v.padding)
+                digest = resolveOption("digest", metadata.algSpecific.supportedDigests, config.rsa.v.digestSpecified, { config.rsa.v.digest }),
+                padding = resolveOption("padding", metadata.algSpecific.supportedPaddings, config.rsa.v.paddingSpecified, { config.rsa.v.padding })
```
JesusMcCloud, line 349: Use constants instead of inline strings

JesusMcCloud (changes requested):
minor nitpicks

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/hazmat/InternalsAccessors.kt
JesusMcCloud: @iaik-jheher sorry, i was off a line. is it still potentially unavailable?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
JesusMcCloud: also;: generics to further narrow down the type of signature would be nice (e.g. to do some in-between processing and make sure every signature is rawByteEncodable, etc)

supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
iaik-jheher: why? it's part of the exception message, we don't usually pull debug messages from constants

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
JesusMcCloud: then i lose the fact its ternary. it just needs to <T: CryptoSignature> for the success case

iaik-jheher: but at that point you also need `transform`, `map`, `mapCatching`, ...

README.md
```diff
@@ -99,6 +99,83 @@ material._
...
+val sha384Signer = key.getSigner { rsa { digest = Digests.SHA384 } }.getOrThrow()
+```
+
+The instances can be configured using the configuration DSL.
+Any unspecified parameters use sensible, secure defaults.
+
+#### Platform Signers
+
+On Android and iOS, signers using the systems' secure key storage can be retrieved.
+To do this, use `PlatformSigningProvider` (in common code), or interact with `AndroidKeystoreProvider`/`IosKeychainProvider` (in platform-specific code).
```
JesusMcCloud, line 126: JKS signer

README.md
```diff
@@ -99,6 +99,83 @@ material._
...
+For example, creating an elliptic-curve key over P256, stored in secure hardware, and with key attestation using a random challenge provided by your server, might be done like this:
+```kotlin
+val serverChallenge: ByteArray = TODO("This was unpredictably chosen by your server.")
+PlatformSigningProvider.createSigningKey(alias = "Swordfish") {
+  ec {
+    // you don't even need to specify the curve (P256 is the default) but we'll do it for demonstration purposes
+    curve = ECCurve.SECP_256_R_1
+    // you could specify the supported digests explicity - if you do not, the curve's native digest (for P256, this is SHA256) is supported
+  }
+  hardware {
```
JesusMcCloud, line 140: We need at least a link to fully documented API docs page for all platforms here

JesusMcCloud (changes requested):
README needs small additions

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/hazmat/InternalsAccessors.kt
iaik-jheher: it is not, comment was amended

supreme/src/androidMain/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProvider.kt
iaik-jheher: is removed

README.md
iaik-jheher: JKS signer is not a `PlatformSigningProvider` right now

README.md
iaik-jheher: done

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
iaik-jheher: can't, build breaks on ios

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
iaik-jheher: see `.modify`

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/Throwables.kt
iaik-jheher: see updated exception hierarchy

supreme/src/androidInstrumentedTest/kotlin/at/asitplus/signum/supreme/os/AndroidKeyStoreProviderTests.kt
iaik-jheher: `should succeed` does not return the contained result. we already use `getOrThrow` consistently for this

README.md
iaik-jheher: added a note

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/SignatureResult.kt
iaik-jheher: added

supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
JesusMcCloud: see #108

PR #95, opened 2024-07-08
Tmp/json key id on init

n0900:
When a JsonWebKey is created from a CryptopublicKey which in turn was created by a keyPair it does not have a keyId. However if we transform this JsonWebKey back to a CryptoPublicKey the identifier gets saved to keyId and if we then transform it back again to JsonWebKey it now has a keyid -> the previous identifier.

As an easy fix I propose this change, however other ideas are welcome.

### DIFF
--- a/datatypes-cose/src/commonMain/kotlin/at/asitplus/crypto/datatypes/cose/CoseKey.kt
+++ b/datatypes-cose/src/commonMain/kotlin/at/asitplus/crypto/datatypes/cose/CoseKey.kt
@@ -11,7 +11,10 @@ import at.asitplus.crypto.datatypes.cose.CoseKeySerializer.CompressedCompoundCos
 import at.asitplus.crypto.datatypes.cose.CoseKeySerializer.UncompressedCompoundCoseKeySerialContainer
 import at.asitplus.crypto.datatypes.cose.io.Base16Strict
 import at.asitplus.crypto.datatypes.cose.io.cborSerializer
+import at.asitplus.crypto.datatypes.io.Base64UrlStrict
+import at.asitplus.crypto.datatypes.io.ByteArrayBase64UrlSerializer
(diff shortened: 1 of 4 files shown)


nodh:
But won't this also add a keyId, when converting from a JWK to a CryptoPublicKey and then back again?

iaik-jheher:
So if I understand you correctly, the conversion from `JsonWebKey` (no keyId) to `CryptoPublicKey` is adding a key ID, where none exists?

If so, why isn't that the issue to fix?

n0900:
> But won't this also add a keyId, when converting from a JWK to a CryptoPublicKey and then back again?

True, this will override a JWK kid if it was null to begin with. Not perfect.

n0900:
> So if I understand you correctly, the conversion from `JsonWebKey` (no keyId) to `CryptoPublicKey` is adding a key ID, where none exists?
> 
> If so, why isn't that the issue to fix?

Because if we fix it like that then no JsonWebKey we generate using our library will ever have a keyID, which might not be ideal

iaik-jheher:
Maybe it should be possible to specify a keyId when converting to `JsonWebKey`, then?

n0900:
> Maybe it should be possible to specify a keyId when converting to `JsonWebKey`, then?

we could lift the additional property from the jsonwebkey to a main property of the CryptoPublicKey itself, what do you think of the recent commit as a base? If this is okay I will extend it for JWK and COSE.

n0900:
Obviously we would then need remove the additional property

iaik-jheher (approved):
@JesusMcCloud @n0900 review please, this should address the issue

nodh (approved):
great!

nodh (approved):

nodh:
Better to squash all commits into one when merging? (-:

PR #96, opened 2024-07-10
Remove identifier from JsonWebKey

nodh:


n0900 (approved):

iaik-jheher:
I wonder if we could maybe have a `SpecializedCryptoPublicKey` interface for JWK, COSE, etc., keys, to parallel the `SpecializedCryptoAlgorithm` interface...

JesusMcCloud (approved):

JesusMcCloud:
> I wonder if we could maybe have a `SpecializedCryptoPublicKey` interface for JWK, COSE, etc., keys, to parallel the `SpecializedCryptoAlgorithm` interface...

These are mostly used as containers for specific use cases and the serialization is the key part. Also: serialization is where JWK is nice and smooth and CoseKey serialization is the what a the perceptive issues of a schizophrenic would look like it if you could translate it to kotlin code. Hence, this might me overkill

iaik-jheher:
Not part of this PR anyway, I'll play around with it.

iaik-jheher (approved):

JesusMcCloud:
> Not part of this PR anyway, I'll play around with it separately.

After all the signers are implemented, i presume ;-)

PR #109, opened 2024-09-04
Remove Swift interop dependency from verifier

iaik-jheher:
Powered by all of the Obj-C interop glue we added for #94.


JesusMcCloud (changes requested):
amend changelog, please

JesusMcCloud (approved):

PR #126, opened 2024-09-16
Fix/implicit tagging

### LINKED ISSUE #13: Revise implicit tagging

JesusMcCloud:
closes #13

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,72 +432,77 @@ DSL, which returns an `Asn1Structure`:
 
 ```kotlin
 Asn1.Sequence {
-    +Tagged(1uL) {
-        +Asn1Primitive(Asn1Element.Tag.BOOL, byteArrayOf(0x00)) //or +Asn1.Bool(false)
-    }
-    +Asn1.Set {
-        +Asn1.Sequence {
-            +Asn1.SetOf {
-                +PrintableString("World")
-                +PrintableString("Hello")
-            }
-            +Asn1.Set {
-                +PrintableString("World")
-                +PrintableString("Hello")
-                +Utf8String("!!!")
-            }
-
-        }
-    }
-    +Asn1.Null()
-
-    +ObjectIdentifier("1.2.603.624.97")
-
-    +Utf8String("Foo")
-    +PrintableString("Bar")
-
-    +Asn1.Set {
-        +Asn1.Int(3)
-        +Asn1.Long(-65789876543L)
-        +Asn1.Bool(false)
-        +Asn1.Bool(true)
-    }
+  +Tagged(1uL) {
+    +Asn1Primitive(Asn1Element.Tag.BOOL, byteArrayOf(0x00)) //or +Asn1.Bool(false)
+  }
+  +Asn1.Set {
     +Asn1.Sequence {
-        +Asn1.Null()
-        +Asn1String.Numeric("12345")
-        +UtcTime(Clock.System.now())
+      +Asn1.SetOf {
+        +PrintableString("World")
+        +PrintableString("Hello")
+      }
+      +Asn1.Set {
+        +PrintableString("World")
+        +PrintableString("Hello")
+        +Utf8String("!!!")
+      }
+
     }
-}
+  }
+  +Asn1.Null()
+
+  +ObjectIdentifier("1.2.603.624.97")
+
+  +(Utf8String("Foo") withImplicitTag (0xCAFEuL withClass TagClass.PRIVATE))
+  +PrintableString("Bar")
+
+  //fake Primitive
+  +(Asn1.Sequence { +Asn1.Int(42) } withImplicitTag (0x5EUL without CONSTRUCTED))
+
+  +Asn1.Set {
+    +Asn1.Int(3)
+    +Asn1.Int(-65789876543L)
+    +Asn1.Bool(false)
+    +Asn1.Bool(true)
+  }
+  +Asn1.Sequence {
+    +Asn1.Null()
+    +Asn1String.Numeric("12345")
+    +UtcTime(Clock.System.now())
+  }
+} withImplicitTag (1337uL withClass TagClass.APPLICATION)
 ```
 
 In accordance with DER-Encoding, this produces the following ASN.1 structure:
 
 ```
-SEQUENCE (8 elem)
-  [1] (1 elem)
-    BOOLEAN false
-  SET (1 elem)
-    SEQUENCE (2 elem)
-      SET (2 elem)
-        PrintableString Hello
-        PrintableString World
-      SET (3 elem)
-        UTF8String !!!
-        PrintableString World
-        PrintableString Hello
-  NULL
-  OBJECT IDENTIFIER 1.2.603.624.97
-  UTF8String Foo
-  PrintableString Bar
-  SET (4 elem)
-    BOOLEAN false
-    BOOLEAN true
-    INTEGER 3
-    INTEGER (36 bit) -65789876543
-  SEQUENCE (3 elem)
+Application 1337 (9 elem)
+
+    [1] (1 elem)
+        BOOLEAN false
+    SET (1 elem)
+        SEQUENCE (2 elem)
+            SET (2 elem)
+                PrintableString World
+                PrintableString Hello
+            SET (3 elem)
+                UTF8String !!!
+                PrintableString World
+                PrintableString Hello
     NULL
-    NumericString 12345
-    UTCTime 2023-10-21 21:14:49 UTC
+    OBJECT IDENTIFIER 1.2.603.624.97
+    Private 51966 (3 byte) Foo
+    PrintableString Bar
+    [94] (3 byte) 02012A
+    SET (4 elem)
+        BOOLEAN false
+        BOOLEAN true
+        INTEGER 3
+        INTEGER (36 bit) -65789876543
+    SEQUENCE (3 elem)
+        NULL
+        NumericString 12345
+        UTCTime 2024-09-16 11:53:51 UTC
 ```
 
 ## Limitations
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ allows for
 
 #### Decoding Values
 
-Various helper functions exist to facilitate decoging the values contained in `Asn1Primitives`, such as `decodeInt()`,
+Various helper functions exist to facilitate decoding the values contained in `Asn1Primitives`, such as `decodeInt()`,
 for example.
 However, anything can be decoded and tagged at will. Therefore, a generic decoding function exists, which has the
 following signature:
@@ -432,7 +432,7 @@ DSL, which returns an `Asn1Structure`:
 
 ```kotlin
 Asn1.Sequence {
-  +Tagged(1uL) {
+  +ExplicitlyTagged(1uL) {
     +Asn1Primitive(Asn1Element.Tag.BOOL, byteArrayOf(0x00)) //or +Asn1.Bool(false)
   }
   +Asn1.Set {
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
@@ -28,33 +28,33 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
     val signature: Asn1Element
 
 
-    /**
-     * Well-defined CryptoSignatures, which can also be encoded to raw bytes, in addition to the DER encoding
-     * specified in the X.509 profile.
-     * RSA Signatures and EC Signatures with a known curve fall into this category.
-     *
-     * **This is the opposite of a [NotRawByteEncodable] signature**
-     */
-    sealed interface RawByteEncodable : CryptoSignature {
         /**
-         * Removes ASN1 Structure and returns the signature value(s) as ByteArray
+         * Well-defined CryptoSignatures, which can also be encoded to raw bytes, in addition to the DER encoding
+         * specified in the X.509 profile.
+         * RSA Signatures and EC Signatures with a known curve fall into this category.
+         *
+         * **This is the opposite of a [NotRawByteEncodable] signature**
          */
-        val rawByteArray: ByteArray
-    }
+        sealed interface RawByteEncodable : CryptoSignature {
+            /**
+             * Removes ASN1 Structure and returns the signature value(s) as ByteArray
+             */
+            val rawByteArray: ByteArray
+        }
 
-    /**
-     * **This is the opposite of a [RawByteEncodable] signature**
-     *
-     * This inverse "non-trait" is required to group [CryptoSignature] subtypes which cannot be encoded into raw byte arrays,
-     * since not all properties required to do so are known. For example, EC signatures parsed from an
-     * [X509Certificate] do not specify a curve. For signatures obtained this way, it is impossible to know
-     * how the components should be padded before encoding it into raw bytes.
-     *
-     * The reason this interface exists, is that it allows for grouping all such signatures in the same manner
-     * as the [RawByteEncodable] ones, to allow for exhaustive `when` clauses
-     *
-     */
-    sealed interface NotRawByteEncodable : CryptoSignature
+        /**
+         * **This is the opposite of a [RawByteEncodable] signature**
+         *
+         * This inverse "non-trait" is required to group [CryptoSignature] subtypes which cannot be encoded into raw byte arrays,
+         * since not all properties required to do so are known. For example, EC signatures parsed from an
+         * [X509Certificate] do not specify a curve. For signatures obtained this way, it is impossible to know
+         * how the components should be padded before encoding it into raw bytes.
+         *
+         * The reason this interface exists, is that it allows for grouping all such signatures in the same manner
+         * as the [RawByteEncodable] ones, to allow for exhaustive `when` clauses
+         *
+         */
+        sealed interface NotRawByteEncodable : CryptoSignature
 
 
     fun encodeToTlvBitString(): Asn1Element
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
@@ -28,33 +28,33 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
     val signature: Asn1Element
 
 
+    /**
+     * Well-defined CryptoSignatures, which can also be encoded to raw bytes, in addition to the DER encoding
+     * specified in the X.509 profile.
+     * RSA Signatures and EC Signatures with a known curve fall into this category.
+     *
+     * **This is the opposite of a [NotRawByteEncodable] signature**
+     */
+    sealed interface RawByteEncodable : CryptoSignature {
         /**
-         * Well-defined CryptoSignatures, which can also be encoded to raw bytes, in addition to the DER encoding
-         * specified in the X.509 profile.
-         * RSA Signatures and EC Signatures with a known curve fall into this category.
-         *
-         * **This is the opposite of a [NotRawByteEncodable] signature**
+         * Removes ASN1 Structure and returns the signature value(s) as ByteArray
          */
-        sealed interface RawByteEncodable : CryptoSignature {
-            /**
-             * Removes ASN1 Structure and returns the signature value(s) as ByteArray
-             */
-            val rawByteArray: ByteArray
-        }
+        val rawByteArray: ByteArray
+    }
 
-        /**
-         * **This is the opposite of a [RawByteEncodable] signature**
-         *
-         * This inverse "non-trait" is required to group [CryptoSignature] subtypes which cannot be encoded into raw byte arrays,
-         * since not all properties required to do so are known. For example, EC signatures parsed from an
-         * [X509Certificate] do not specify a curve. For signatures obtained this way, it is impossible to know
-         * how the components should be padded before encoding it into raw bytes.
-         *
-         * The reason this interface exists, is that it allows for grouping all such signatures in the same manner
-         * as the [RawByteEncodable] ones, to allow for exhaustive `when` clauses
-         *
-         */
-        sealed interface NotRawByteEncodable : CryptoSignature
+    /**
+     * **This is the opposite of a [RawByteEncodable] signature**
+     *
+     * This inverse "non-trait" is required to group [CryptoSignature] subtypes which cannot be encoded into raw byte arrays,
+     * since not all properties required to do so are known. For example, EC signatures parsed from an
+     * [X509Certificate] do not specify a curve. For signatures obtained this way, it is impossible to know
+     * how the components should be padded before encoding it into raw bytes.
+     *
+     * The reason this interface exists, is that it allows for grouping all such signatures in the same manner
+     * as the [RawByteEncodable] ones, to allow for exhaustive `when` clauses
+     *
+     */
+    sealed interface NotRawByteEncodable : CryptoSignature
 
 
     fun encodeToTlvBitString(): Asn1Element
@@ -212,7 +212,7 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
                 decodeFromDer(src.readBitString().rawBytes)
             }
 
-            override fun decodeFromTlv(src: Asn1Element): EC.IndefiniteLength {
+            override fun doDecode(src: Asn1Element): EC.IndefiniteLength {
                 src as Asn1Sequence
                 val r = (src.nextChild() as Asn1Primitive).readBigInteger()
                 val s = (src.nextChild() as Asn1Primitive).readBigInteger()
@@ -222,6 +222,7 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
 
             @Deprecated("use fromRawBytes", ReplaceWith("CryptoSignature.EC.fromRawBytes(input)"))
             operator fun invoke(input: ByteArray): DefiniteLength = fromRawBytes(input)
+
         }
 
     }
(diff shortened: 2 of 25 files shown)


iaik-jheher:
please remove whitespace changes so we can review the actual PR diff

JesusMcCloud:
@iaik-jheher three lines of whitespace? I fail to see the issue

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
```diff
@@ -28,33 +28,33 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
...
 
 
+    /**
+     * Well-defined CryptoSignatures, which can also be encoded to raw bytes, in addition to the DER encoding
+     * specified in the X.509 profile.
+     * RSA Signatures and EC Signatures with a known curve fall into this category.
+     *
+     * **This is the opposite of a [NotRawByteEncodable] signature**
+     */
+    sealed interface RawByteEncodable : CryptoSignature {
```
iaik-jheher, line 38: this entire section appears to do nothing except change the indentation

iaik-jheher: (i did not review the rest of the PR)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -162,8 +220,22 @@ sealed class Asn1Element(
                 return derEncoded
             }
 
+            /**
+             * Convenience helper to easily construct implicitly tagged elements.
+             * Shorthand for `Tag(tagValue, constructed=false, tagClass=TagClass.CONTEXT_SPECIFIC)
+             */
+            fun ULong.toImplicitTag() =
```
iaik-jheher, line 227: Is this not duplicating the (more explicit, pun not intended) `long withClass CONTEXT_SPECIFIC`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -196,7 +267,7 @@ sealed class Asn1Element(
 
         val isConstructed by lazy { encodedTag.first().toUByte().isConstructed() }
 
-        val isExplicitlyTagged by lazy { isConstructed && tagClass == TagClass.CONTEXT_SPECIFIC }
+        internal val isExplicitlyTagged by lazy { isConstructed && tagClass == TagClass.CONTEXT_SPECIFIC }
```
iaik-jheher, line 270: this feels like it should be a getter

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -196,7 +267,7 @@ sealed class Asn1Element(
 
         val isConstructed by lazy { encodedTag.first().toUByte().isConstructed() }
```
iaik-jheher, line 268: this also feels like it should be a getter

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -233,6 +304,50 @@ sealed class Asn1Element(
...
+         * creates a new Tag from this object, overriding the class. Useful for implicitTagging (see [Asn1Structure.withImplicitTag])
+         */
+        infix fun withClass(tagClass: TagClass) =
+            Tag(this.tagValue, tagClass = tagClass, constructed = this.isConstructed)
+
+        /**
+         * creates a new Tag from this object, negating the passed property. Useful for implicitTagging (see [Asn1Structure.withImplicitTag]).
+         * This is a NOOP for tag that don't have this bit set.
+         */
+        infix fun without(negated: TagProperty): Tag = when (negated) {
```
iaik-jheher, line 318: did i miss it, or is `with` missing?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -321,32 +475,36 @@ class Asn1Sequence internal constructor(children: List<Asn1Element>) :
...
-    tag: Tag, children: List<Asn1Element>
+    tag: Tag, children: List<Asn1Element>, sort: Boolean
 ) :
     Asn1Structure(tag, children) {
     /**
      * ASN.1 CONSTRUCTED with custom tag
      * @param children the elements to put into this sequence
      * @param tag the custom tag to use
      * @param tagClass the tag class to use for this custom tag. defaults to [TagClass.UNIVERSAL]
+     * @param sort whether to sort the passed childr nodes. defautls to false
```
iaik-jheher, line 486: spelling: child, defaults

iaik-jheher (changes requested):

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -127,6 +128,63 @@ sealed class Asn1Element(
         return result
     }
 
+
+    /**
+     * Creates a new implicitly tagged ASN.1 Element from this ASN.1 Element.
+     * NOTE: The [TagClass] of the provided [tag] will be used! If you want the result to have [TagClass.CONTEXT_SPECIFIC],
+     * also invoke `tag withClass TagClass.CONTEXT_SPECIFIC`!. If a CONSTRUCTED Tag is applied to an ASN.1 Primitive,
```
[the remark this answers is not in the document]
JesusMcCloud: the first point i get, the second one, not. there are exactly four legal states for those bits and no extension is possible

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -127,6 +128,63 @@ sealed class Asn1Element(
...
+        }
+
+        is Asn1Primitive -> Asn1Primitive(tag without CONSTRUCTED, content)
+    }
+
+    /**
+     * Creates a new implicitly tagged ASN.1 Element from this ASN.1 Element.
+     * Sets the class of the resulting structure to [TagClass.CONTEXT_SPECIFIC]
+     */
+    inline infix fun withImplicitTag(tagValue: ULong) = when (this) {
```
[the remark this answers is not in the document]
JesusMcCloud: no, it should not, because this one uses a pre-set class and the other does not (and the Docs also specify that)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: it is, but you will need this all over the place when dealing with X.509

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
iaik-jheher: the second one is purely about making it less verbose (`tag withClass CONTEXT_SPECIFIC` instead of `tag withClass TagClass.CONTEXT_SPECIFIC`)

by doing
```kotlin
sealed interface TagClass {
  companion object {
    val CONTEXT_SPECIFIC = ::CONTEXT_SPECIFIC
  }
}
data object CONTEXT_SPECIFIC: TagClass
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: why would you invite the creation of illegal ASN.1 structures that don't parse?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: then maybe a `Asn1.Tag(ULong)` constructor?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: ah, right

should `Tagged(42)` maybe be `Constructed(tag = 42)`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: i think we should agree to disagree on this point. four values, no extensions ever possible, zero complexity in those values -> it's an enum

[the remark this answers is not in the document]
iaik-jheher: my example has no added complexity 🤷 it just makes the code more readable/less verbose, same argument as something like `ULong.asImplicitTag` over `ULong withClass CONTEXT_SPECIFIC`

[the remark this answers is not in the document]
JesusMcCloud: neither two classes with the same name or one class which should behave like an enum, but does not is not my definition of "no added complexity", when i could just use a static import to get the same result in my code (see my last commit)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: that would cause more confusion: is it implicit? should I use that when invoking Asn1Tagged? etc…

I do get that both structures and primitives may be tagged implicitly.
Still, this will be used and the function name combined with the API docs leaves little to be desired.
Build some X.509 structures. implicitly tagged primitives are just so common, that this shorthand is warranted.

iaik-jheher: `Asn1.ImplicitTag(ULong, tagClass = TagClass.CONTEXT_SPECIFIC)`?

iaik-jheher: (i'm especially concerned about polluting the IDE auto-complete with "toImplicitTag", where it is not immediately clear what context "implicit tag" is even referring to... and if you call it "toImplicitAsn1Tag" the constructor is the same length)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: please provide a more comprehensive example, as I have a hard time wrapping my head around this suggestion (and maybe check [RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280) and how your proposal aligns with the syntax and lingo there)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: fixed. thanks!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: > `Asn1.ImplicitTag(ULong, tagClass = TagClass.CONTEXT_SPECIFIC)`?
> 
> `42u.toImplicitTag()` is hardly shorter than `Asn1.ImplicitTag(42u)`

I like that! done (also for explicit tag)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: done and ready for review

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/AlternativeNames.kt
```diff
@@ -132,13 +131,13 @@ private constructor(private val extensions: List<Asn1Element>) {
...
-    val iPAddress = 7uL.toImplicitTag()
-    val registeredID = 8uL.toImplicitTag()
+    val otherName = Asn1.ImplicitTag(0uL)
+    val rfc822Name = Asn1.ImplicitTag(1uL)
+    val dNSName = Asn1.ImplicitTag(2uL)
+    val x400Address = Asn1.ImplicitTag(3uL)
+    val directoryName = Asn1.ImplicitTag(4uL)
+    val ediPartyName = Asn1.ImplicitTag(5uL)
+    val uniformResourceIdentifier = Asn1.ImplicitTag(6uL)
+    val iPAddress = Asn1.ImplicitTag(7uL)
```
iaik-jheher, line 141: spelling: `ipAddress`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/AlternativeNames.kt
```diff
@@ -132,13 +131,13 @@ private constructor(private val extensions: List<Asn1Element>) {
...
-    val dNSName = 2uL.toImplicitTag()
-    val x400Address = 3uL.toImplicitTag()
-    val directoryName = 4uL.toImplicitTag()
-    val ediPartyName = 5uL.toImplicitTag()
-    val uniformResourceIdentifier = 6uL.toImplicitTag()
-    val iPAddress = 7uL.toImplicitTag()
-    val registeredID = 8uL.toImplicitTag()
+    val otherName = Asn1.ImplicitTag(0uL)
+    val rfc822Name = Asn1.ImplicitTag(1uL)
+    val dNSName = Asn1.ImplicitTag(2uL)
```
iaik-jheher, line 136: spelling: `dnsName`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: should there be an `ExplicitlyTagged(tag: Asn1.Tag)` overload that checks if the tag is explicit?

JesusMcCloud: It would probably make more sense, to add `Structure(tag:Tag)`, letting you specify whatever you want, that returns a custom structure

JesusMcCloud: `Sequence{…} withImplicitTag` does that and does not get in your way with tag-altering checks

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/AlternativeNames.kt
JesusMcCloud: this is intentional: it uses the same off-casing as the RFC that specifies it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: thing is, if you override anything but the numeric tag value, it is not an explicit tag any more, so it really does not make sense, because you'll end up with a custom structure

iaik-jheher: what if i want to take the (explicit) tag of some structure and create a new structure with the same (explicit) tag?

iaik-jheher: I'm thinking `fun ExplicitlyTagged(tag: Asn1.Tag, foo: ()->whatever) { require(tag.isExplicitTag); return ExplicitlyTagged(tag.tagValue, foo); }`

JesusMcCloud: > what if i want to take the (explicit) tag of some structure and create a new structure with the same (explicit) tag?

`Asn1.ExplicitTag(oldSctrucuter.tag.tagValue) {…}`

Again, the only variable for this DSL invocation is the numeric tag value, all other bits are fixed, so it makes no sense to allow passing other properties, just to cause confusion and invite user errors

iaik-jheher:
resolve conflicts please

README.md
```diff
@@ -432,72 +432,77 @@ DSL, which returns an `Asn1Structure`:
...
-        +Asn1.Bool(true)
-    }
-    +Asn1.Sequence {
-        +Asn1.Null()
-        +Asn1String.Numeric("12345")
-        +UtcTime(Clock.System.now())
-    }
-}
+  //fake Primitive
+  +(Asn1.Sequence { +Asn1.Int(42) } withImplicitTag (0x5EUL without CONSTRUCTED))
```
iaik-jheher, line 460: can we maybe use a pre-defined constructed tag (do we have this?) to give a more realistic example?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -181,6 +182,54 @@ sealed class Asn1Element(
...
+
+    /**
+     * Creates a new implicitly tagged ASN.1 Element from this ASN.1 Element.
+     * NOTE: The [TagClass] of the provided [tag] will be used! If you want the result to have [TagClass.CONTEXT_SPECIFIC],
+     * use `element withImplicitTag (tag withClass TagClass.CONTEXT_SPECIFIC)`!. If a CONSTRUCTED Tag is applied to an ASN.1 Primitive,
+     * the CONSTRUCTED bit is overridden and set to zero.
+     */
+    inline infix fun withImplicitTag(tag: Tag): Asn1Element = when (this) {
+        is Asn1Structure -> {
+            if (tag.isConstructed) Asn1CustomStructure(
```
iaik-jheher, line 194: indentation here is weird

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -306,9 +398,16 @@ object Asn1EncodableSerializer : KSerializer<Asn1Element> {
...
+sealed class Asn1Structure(tag: Tag, children: List<Asn1Element>?, sort: Boolean = false) :
+    Asn1Element(TLV(tag, byteArrayOf()), if (!sort) children else children?.sortedBy { it.tag }) {
 
+    /**
+     * Whether this structure sorts child nodes or keeps them as-is.
+     * This **should** be true for SET and SET OF, but is set to false for SET and SET OF elements parsed
+     * from DER-encoded structures, because this has a chance of altering the structure for non-conforming DER-encoded
+     * structures.
+     */
+    val isSorted: Boolean = sort
```
iaik-jheher, line 410: why is this not in the constructor then

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -306,9 +398,16 @@ object Asn1EncodableSerializer : KSerializer<Asn1Element> {
...
  */
-sealed class Asn1Structure(tag: Tag, children: List<Asn1Element>?) :
-    Asn1Element(TLV(tag, byteArrayOf()), children) {
+sealed class Asn1Structure(tag: Tag, children: List<Asn1Element>?, sort: Boolean = false) :
+    Asn1Element(TLV(tag, byteArrayOf()), if (!sort) children else children?.sortedBy { it.tag }) {
 
+    /**
+     * Whether this structure sorts child nodes or keeps them as-is.
+     * This **should** be true for SET and SET OF, but is set to false for SET and SET OF elements parsed
+     * from DER-encoded structures, because this has a chance of altering the structure for non-conforming DER-encoded
```
iaik-jheher, line 407: name & shame some examples 😈

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Encodable.kt
```diff
@@ -43,88 +43,90 @@ interface Asn1Encodable<A : Asn1Element> {
...
+        val expected = tagOverride ?: return
+        if (src.tag != expected)
+            throw Asn1TagMismatchException(expected, src.tag)
+    }
 
     /**
      * Exception-free version of [decodeFromTlv]
      */
-    fun decodeFromTlvOrNull(src: Asn1Primitive, tagOverride: Asn1Element.Tag?) =
+    fun decodeFromTlvOrNull(src: A, tagOverride: Asn1Element.Tag? = null) =
```
iaik-jheher, line 104: wasn't the plan to get rid of all of the tag override parameters? do we still need these?

iaik-jheher (changes requested):

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: one comes to mind, but i am confident neither of us can even fathom how much garbage is out there. 

For the record: Apple is one offender that comes to mind

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Encodable.kt
JesusMcCloud: ??? you absolutely need this to decode while verifying implicit tags.

iaik-jheher: ah, it's not really a "tag override", it's more of an "asserted tag"? is this the right name?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: i meant give it as an example in the docs for why this is needed in practice, otherwise it's unintuitive why this would account for non-compliant code

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: good catch!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: i fail to see it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: the comment already explains why this counter-intuitive flag is required. i am against pointing fingers in the docs here, because our implementation is also not feature-complete and allows for generating non-compliant structures, if we apply the SET tag to custom structures

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Encodable.kt
JesusMcCloud: you are right. it only checks if present. fixed!

README.md
JesusMcCloud: A sequence has a predefined tag of 16 and has the constructed tag set. To me, this is a prime example of where how to implicitly tag and override the constructed bit

iaik-jheher (approved):
approved conditional on CI build

Issue #121, opened 2024-09-16
ASN.1: Parser API changes


JesusMcCloud:
Currently, the parser is eager to consume all bytes, when trying to parse an ASN.1 Structure.
This makes little sense, because only a single toplevel structure is allowed. If, for example, some input (which can be perfectly legal in a couple of use cases) starts with a valid ASN.1 structure, followed by other data (possibly even additional valid ASN.1 structures), parsing will fail.

Hence, the toplevel `Asn1Element.parse` and `Asn1Element.decodeFromDerHexString` should only try to parse the very first ASN.1 Structure and also return all non-parsed trailing bytes.

PR #128, opened 2024-09-16
Consume only the first ASN.1 Element when parsing

JesusMcCloud:
addresses #121.
this is a stop-gap solution until we get to adopting kotlinx.io


JesusMcCloud:
superseded by #130

PR #153, opened 2024-10-02
Add Asn1EncodableBase64Serializer

n0900:
Adds Asn1EncodableBase64Serializer which is used in RQES serialization


JesusMcCloud:
Kotlinx-serializable is used for debugging ASN.1 structures, not to be stable. you should rather annotate the property where you require this behaviour with this serializer and add the serializer to the project where this is required.

Feel free to reopen with convincing arguments ;-)

PR #155, opened 2024-10-04
Migrate JCA signing from DER encoding to P1363 encoding

iaik-jheher:
DER-encoded ECDSA signatures are an ASN.1 structure, with variable length integer representation. P1363-encoded ECDSA signatures are fixed-length concatenated integers with no additional structure.

This means that P1363 -> DER is easy, while DER -> P1363 is impossible without additional information (such as the curve used).
We currently use the `RawByteEncodable` marker interface to indicate whether a given (EC) signature can be encoded to P1363.

The JCA extensions in Indispensable currently produce and consume DER signatures. This is awkward, since it means that `CryptoSignature.parseFromJCA` cannot return `RawByteEncodable` signatures -- not even in the RSA case!

This PR fixes this. It declares the individual parts of the signing process as hazmat (since they are encoding dependent) and switches them to use P1363. The stable API going forward for signing ops using the JCA will be `.signWithJCA`, which guarantees that the signature object and signature byte decoding use the same encoding (whatever that may be on any given platform).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 * Change CSR to take an actual `CryptoSignature` instead of a ByteArray
 * Introduce shorthand to create CSR from TbsCSR
 * Introduce shorthand to create certificate from TbsCertificate
+* Add `SignatureAlgorithm.signWithJCA` and `SignatureAlgorithm.verifyWithJCA` helpers.
+* Deprecate `SignatureAlgorithm.getJCASignatureInstance`, `CryptoSignature.parseFromJCA` etc as hazmat.
 
 
 ### 3.9.0 (Supreme 0.4.0)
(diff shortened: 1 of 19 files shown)


JesusMcCloud:
so… do you want me to add a dedicated android target? if you fear duplicated code, we can just add a `jvmAndroidMain` source folder and hook it into the JVM and Android source sets, so everything **but** the culprit can be shared and we require no code duplication

JesusMcCloud:
@iaik-jheher so… what do you want me to do about this?

iaik-jheher:
not sure, I might even abandon it altogether, since it "just" deals with glue code... I though this would be a quick and free fix

iaik-jheher:
the extra source set hierarchy for android/java might be a good idea regardless, just so we can get rid of the "isAndroid" classpath checking hack

JesusMcCloud:
we have the hierarchy now, so I'm closing this

PR #158, opened 2024-10-07
Fix/oid

JesusMcCloud:
Revamps OIDs to be based on BigIntegers and directly supports 2.25.<UUID int representation> OIDs.

Yes, this halves OID performance, but the thing is: we need all validations while parsing, so basing everything on Bytes is nice for creation, but we'd need to move all parsing logic into the primary constructor.   
Hence, using BigInts as the foundation may be a bit slower, but creating and parsing remains symmetric.
The performance difference would be 1/3 AT BEST, realistically, it would probably more like 15-20% at the cost of harder-to-comprehend code

### DIFF
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
@@ -1,8 +1,9 @@
 package at.asitplus.signum.indispensable.asn1
 
 import at.asitplus.signum.indispensable.asn1.encoding.decode
-import at.asitplus.signum.indispensable.asn1.encoding.decodeAsn1VarUInt
+import at.asitplus.signum.indispensable.asn1.encoding.decodeAsn1VarBigInt
 import at.asitplus.signum.indispensable.asn1.encoding.toAsn1VarInt
+import com.ionspin.kotlin.bignum.integer.BigInteger
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.Transient
@@ -10,6 +11,8 @@ import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
+
+private val BIGINT_40 = BigInteger.fromUByte(40u)
 
 /**
  * ASN.1 OBJECT IDENTIFIER featuring the most cursed encoding of numbers known to man, which probably surfaced due to an ungodly combination
@@ -19,12 +22,12 @@ import kotlinx.serialization.encoding.Encoder
  * @throws Asn1Exception if less than two nodes are supplied, the first node is >2 or the second node is >39
  */
 @Serializable(with = ObjectIdSerializer::class)
-class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vararg val nodes: UInt) :
+class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vararg val nodes: BigInteger) :
     Asn1Encodable<Asn1Primitive> {
 
     init {
         if (nodes.size < 2) throw Asn1StructuralException("at least two nodes required!")
-        if (nodes[0] * 40u > UByte.MAX_VALUE.toUInt()) throw Asn1Exception("first node too lage!")
+        if ((nodes[0] * BIGINT_40) > UByte.MAX_VALUE.toUInt()) throw Asn1Exception("first node too lage!")
         //TODO more sanity checks
 
         if (nodes.first() > 2u) throw Asn1Exception("OID must start with either 1 or 2")
@@ -32,9 +35,17 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
     }
 
     /**
+     * @param nodes OID Tree nodes passed in order (e.g. 1u, 2u, 96u, …)
+     * @throws Asn1Exception if less than two nodes are supplied, the first node is >2 or the second node is >39
+     */
+    constructor(vararg ints: UInt) : this(*(ints.map { BigInteger.fromUInt(it) }.toTypedArray()))
+
+
+    /**
      * @param oid in human-readable format (e.g. "1.2.96")
      */
-    constructor(oid: String) : this(*(oid.split(if (oid.contains('.')) '.' else ' ')).map { it.toUInt() }.toUIntArray())
+    constructor(oid: String) : this(*(oid.split(if (oid.contains('.')) '.' else ' ')).map { BigInteger.parseString(it) }
+        .toTypedArray())
 
     /**
      * @return human-readable format (e.g. "1.2.96")
@@ -48,7 +59,7 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
     }
 
     override fun hashCode(): Int {
-        return nodes.hashCode()
+        return bytes.contentHashCode()
     }
 
 
@@ -58,7 +69,7 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
     val bytes: ByteArray by lazy {
         nodes.slice(2..<nodes.size).map { it.toAsn1VarInt() }.fold(
             byteArrayOf(
-                (nodes[0] * 40u + nodes[1]).toUByte().toByte()
+                (nodes[0] * BIGINT_40 + nodes[1]).ubyteValue(exactRequired = true).toByte()
             )
         ) { acc, bytes -> acc + bytes }
     }
@@ -92,16 +103,16 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
             if (rawValue.isEmpty()) throw Asn1Exception("Empty OIDs are not supported")
             val (first, second) =
                 if (rawValue[0] >= 80) {
-                    2u to rawValue[0].toUByte() - 80u
+                    BigInteger.fromUByte(2u) to BigInteger.fromUInt(rawValue[0].toUByte() - 80u)
                 } else {
-                    rawValue[0].toUByte() / 40u to rawValue[0].toUByte() % 40u
+                    BigInteger.fromUInt(rawValue[0].toUByte() / 40u) to BigInteger.fromUInt(rawValue[0].toUByte() % 40u)
                 }
 
             var index = 1
             val collected = mutableListOf(first, second)
             while (index < rawValue.size) {
                 if (rawValue[index] >= 0) {
-                    collected += rawValue[index].toUInt()
+                    collected += BigInteger.fromUInt(rawValue[index].toUInt())
                     index++
                 } else {
                     val currentNode = mutableListOf<Byte>()
@@ -111,10 +122,10 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
                     }
                     currentNode += rawValue[index]
                     index++
-                    collected += currentNode.decodeAsn1VarUInt().first
+                    collected += currentNode.decodeAsn1VarBigInt().first
                 }
             }
-            return ObjectIdentifier(*collected.toUIntArray())
+            return ObjectIdentifier(*collected.toTypedArray())
         }
     }
 }
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
@@ -48,6 +48,7 @@ class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
             }else {
                 if (get(1) > 47u) throw Asn1Exception("Second segment must be <48")
             }
+            forEach { if (it.isNegative) throw Asn1Exception("Negative Number encountered: $it") }
         }
     }
 
@@ -169,27 +170,13 @@ class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
         fun parse(rawValue: ByteArray): ObjectIdentifier = ObjectIdentifier(rawValue)
 
         private fun UIntArray.toOidBytes(): ByteArray {
-            if (size < 2) throw Asn1StructuralException("at least two nodes required!")
-            if (first() > 2u) throw Asn1Exception("OID top-level arc can only be number 0, 1 or 2")
-            if(first()<2u) {
-                if (get(1) > 39u) throw Asn1Exception("Second segment must be <40")
-            }else {
-                if (get(1) > 47u) throw Asn1Exception("Second segment must be <48")
-            }
             return slice(2..<size).map { it.toAsn1VarInt() }.fold(
                 byteArrayOf((first() * 40u + get(1)).toUByte().toByte())
             ) { acc, bytes -> acc + bytes }
         }
 
         private fun List<out BigInteger>.toOidBytes(): ByteArray {
-            if (size < 2) throw Asn1StructuralException("at least two nodes required!")
-            if (first() > 2u) throw Asn1Exception("OID top-level arc can only be number 0, 1 or 2")
-            if(first()<2u) {
-                if (get(1) > 39u) throw Asn1Exception("Second segment must be <40")
-            }else {
-                if (get(1) > 47u) throw Asn1Exception("Second segment must be <48")
-            }
-            return slice(2..<size).map { if (it.isNegative) throw Asn1Exception("Negative Number encountered: $it") else it.toAsn1VarInt() }
+            return slice(2..<size).map { it.toAsn1VarInt() }
                 .fold(
                     byteArrayOf((first().intValue() * 40 + get(1).intValue()).toUByte().toByte())
                 ) { acc, bytes -> acc + bytes }
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
@@ -25,14 +25,14 @@ private val BIGINT_40 = BigInteger.fromUByte(40u)
  * @throws Asn1Exception if less than two nodes are supplied, the first node is >2 or the second node is >39
  */
 @Serializable(with = ObjectIdSerializer::class)
-class ObjectIdentifier @Throws(Asn1Exception::class) constructor(
+class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
     val bytes: ByteArray,
     @Transient private val _nodes: List<BigInteger>? = null, dontVerify: Boolean = false
 ) :
     Asn1Encodable<Asn1Primitive> {
 
     init {
-        if (_nodes == null || dontVerify) {
+        if (_nodes == null || !dontVerify) {
             //Verify that everything can be parsed into nodes
             if (bytes.isEmpty()) throw Asn1Exception("Empty OIDs are not supported")
             var index = 1
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
@@ -3,7 +3,9 @@ package at.asitplus.signum.indispensable.asn1
 import at.asitplus.signum.indispensable.asn1.encoding.decode
 import at.asitplus.signum.indispensable.asn1.encoding.decodeAsn1VarBigInt
 import at.asitplus.signum.indispensable.asn1.encoding.toAsn1VarInt
+import at.asitplus.signum.indispensable.asn1.encoding.toBigInteger
 import com.ionspin.kotlin.bignum.integer.BigInteger
+import com.ionspin.kotlin.bignum.integer.Sign
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.Transient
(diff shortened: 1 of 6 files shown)


iaik-jheher:
have you benchmarked this? `BigInteger` is heavyweight enough that you're probably knocking out a bunch of locality optimizations on JVM with this

JesusMcCloud:
I benchmarked only the JVM, not on native and the performance loss for encoding and decoding combined was 50%.

An alternative would be to only work on bytes, i.e. have neither UInt nor BigInteger nodes, but just the encoded bytes as the foundation of the OID class. This would remove all the BigInteger overhead from creation, if, for example, you pass only UInts to the constructor (which would be 99.999% of the time). When decoding from DER, or encoding to String, most of the BigInteger overhead would come back though, because we'd need to verify that the bytes are indeed a valid representation of an OID. You could, of course optimize the code when parsing to not ever create a BigInteger instance. Then only `toString` would be slow but we'd have more almost code duplicates. Probably still a viable approach, but it would increase the complexity of the code.

JesusMcCloud:
Also nevermind the failing tests. Seems like it just so happens that the random bytes accidentally made up an ASN.1 structure.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+    }
+
+    /**
+     * Creates an OID in the 2.25 subtree that requires no formal registration.
+     * E.g. the UUID `550e8400-e29b-41d4-a716-446655440000` results in the OID
+     * `2.25.113059749145936325402354257176981405696`
+     */
+    @OptIn(ExperimentalUuidApi::class)
+    constructor(uuid: Uuid) : this(
+        byteArrayOf((2 * 40 + 25).toUByte().toByte(), *uuid.toBigInteger().toAsn1VarInt()), dontVerify = true
```
iaik-jheher, line 116:
```suggestion
        bytes = byteArrayOf((2 * 40 + 25).toUByte().toByte(), *uuid.toBigInteger().toAsn1VarInt()),
        dontVerify = true
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+    constructor(vararg nodes: UInt) : this(
+        nodes.toOidBytes(),
+        dontVerify = true
+    )
+
+    /**
+     * @param nodes OID Tree nodes passed in order (e.g. 1, 2, 96, …)
+     * @throws Asn1Exception if less than two nodes are supplied, the first node is >2, the second node is >39 or any node is negative
+     */
+    constructor(vararg nodes: BigInteger) : this(bytes = null, nodes = nodes.asList())
```
iaik-jheher, line 132:
```suggestion
    constructor(vararg nodes: BigInteger) : this(nodes = nodes.asList())
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+
     /**
      * @param oid in human-readable format (e.g. "1.2.96")
      */
-    constructor(oid: String) : this(*(oid.split(if (oid.contains('.')) '.' else ' ')).map { it.toUInt() }.toUIntArray())
+    constructor(oid: String) : this(*(oid.split(if (oid.contains('.')) '.' else ' ')).map { BigInteger.parseString(it) }
+        .toTypedArray())
```
iaik-jheher, line 138: this has a hidden feature which allows splitting on whitespace; either document it or remove it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -88,33 +180,37 @@ class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vara
...
-                    val currentNode = mutableListOf<Byte>()
-                    while (rawValue[index] < 0) {
-                        currentNode += rawValue[index] //+= parsed
-                        index++
-                    }
-                    currentNode += rawValue[index]
-                    index++
-                    collected += currentNode.decodeAsn1VarUInt().first
-                }
+        fun parse(rawValue: ByteArray): ObjectIdentifier = ObjectIdentifier(rawValue)
```
iaik-jheher, line 183: is there a reason why this is a static member while the others are secondary constructors? seems weird

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+        }
+    }
+
+
+    /**
+     * Cursed encoding of OID nodes. A sacrifice of pristine numbers requested by past gods of the netherrealm.
+     * Lazily evaluated
+     */
+    val bytes: ByteArray by if (bytes != null) lazyOf(bytes) else lazy {
+        nodes ?: throw Asn1Exception("Empty nodes are not supported")
```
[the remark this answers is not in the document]
JesusMcCloud: how? it's the nullable constructor parameter

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+        nodes.toOidBytes()
     }
 
+    /**
+     * Lazily evaluated list of OID nodes (e.g. `[1, 2, 35, 4654]`)
+     */
+    val nodes by if (nodes != null) lazyOf(nodes) else lazy {
+        bytes ?: throw Asn1Exception("Empty nodes are not supported")
```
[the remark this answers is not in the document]
JesusMcCloud: how? it's the nullable constructor parameter

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
  * @throws Asn1Exception if less than two nodes are supplied, the first node is >2 or the second node is >39
  */
 @Serializable(with = ObjectIdSerializer::class)
-class ObjectIdentifier @Throws(Asn1Exception::class) constructor(@Transient vararg val nodes: UInt) :
+class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
+    bytes: ByteArray?,
+    nodes: List<BigInteger>? = null,
+    dontVerify: Boolean = false
```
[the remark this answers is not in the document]
JesusMcCloud: then the defaults would be definitely illegal?!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: usually, when you construct OIDs you are not handling bytes, you are either passing strings or a list of nodes. Here you are parsing (more like validating), encoded bytes. this is in line with the rest of the API

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
-
     init {
-        if (nodes.size < 2) throw Asn1StructuralException("at least two nodes required!")
-        if (nodes[0] * 40u > UByte.MAX_VALUE.toUInt()) throw Asn1Exception("first node too lage!")
-        //TODO more sanity checks
 
-        if (nodes.first() > 2u) throw Asn1Exception("OID must start with either 1 or 2")
-        if (nodes[1] > 39u) throw Asn1Exception("Second segment must be <40")
+        //we're not even declaring these, since this is an implementation error on our end
+        if (bytes == null && nodes == null)
```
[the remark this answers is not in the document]
JesusMcCloud: ?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+            throw IllegalArgumentException("either nodes or bytes required")
+        if (bytes?.isEmpty() == true && nodes?.isEmpty() == true)
```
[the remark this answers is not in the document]
JesusMcCloud: that is actually true with the latest changes. I am starting to reconsider whether this whole mess is worth the performance boost.  and it is a mess…

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+                    }
+                    currentNode += bytes[index]
+                    index++
+                    val consumed = currentNode.iterator().consumeVarIntEncoded()
+                    @OptIn(ExperimentalStdlibApi::class)
+                    if (consumed != currentNode) throw Asn1Exception(
+                        "Trailing bytes in OID Node ${
+                            currentNode.toByteArray().toHexString(HexFormat.UpperCase)
+                        }"
+                    )
```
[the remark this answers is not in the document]
JesusMcCloud: The whole verification is bonkers, if we're not parsing! every byte string can be interpreted as concatenated uvarints!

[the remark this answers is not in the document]
JesusMcCloud: getting rid of this made constructing oids a NOOP, busines-logic-wise, so this got us a 1000% (not a typo!) performance gain

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+                }
+            }
+        }
+    }
+
+
+    /**
+     * Cursed encoding of OID nodes. A sacrifice of pristine numbers requested by past gods of the netherrealm.
```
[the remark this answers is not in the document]
JesusMcCloud: You're looking for a quote from Shakespeare's Hamlet:
> more matter, with less art!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: documented

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: nope. making this work would require the default constructor parameters to represent an illegal state. not a fan of this strategy

JesusMcCloud:
I've just revamped the validations. at the expense of some performance it now happens in one place and is exactly aligned with Bouncy castle behaviour. Also: Everything is not thoroughly tested and there should not be any loopholes any more

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
iaik-jheher: oh, i thought this was referring to the member; why not `this.bytes` explicitly? either way this is guaranteed to be non-null

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
iaik-jheher: both parameters being non-null should also be an error case; secondary constructors should always specify exactly one

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
iaik-jheher: reconsider this in light of the latest changes, where there is no parsing/validating involved?

iaik-jheher:
Now I am confused, I thought you concluded the validation was always passing?

JesusMcCloud:
> Now I am confused, I thought you concluded the validation was always passing?

The first two nodes require validation and when passing bigints we need to make sure that every component is positive

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
JesusMcCloud: cleaned up now

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
JesusMcCloud: using this.nodes now

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
JesusMcCloud: using this.bytes now

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,135 @@ import kotlinx.serialization.encoding.Encoder
...
+    )
+
+    /**
+     * @param nodes OID Tree nodes passed in order (e.g. 1u, 2u, 96u, …)
+     * @throws Asn1Exception if less than two nodes are supplied, the first node is >2 or the second node is >39
+     */
+    constructor(vararg nodes: UInt) : this(
+        nodes.toOidBytes(),
+        dontVerify = true
+    )
```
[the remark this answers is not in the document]
JesusMcCloud: every parameter is now explicitly declared

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: every parameter is now explicitly declared

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: did the opposite, now requiring explicitly declaring nulls either way

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: see my comment below. the first node requires validation

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,125 @@ import kotlinx.serialization.encoding.Encoder
...
-        if (nodes[0] * 40u > UByte.MAX_VALUE.toUInt()) throw Asn1Exception("first node too lage!")
-        //TODO more sanity checks
+        if ((bytes == null) && (nodes == null)) {
+            //we're not even declaring this, since this is an implementation error on our end
+            throw IllegalArgumentException("either nodes or bytes required")
+        }
+        if (bytes?.isEmpty() == true || nodes?.isEmpty() == true)
+            throw Asn1Exception("Empty OIDs are not supported")
+
+        bytes?.apply {
```
iaik-jheher, line 40: maybe move this to the `lazyOf` initializer?

`lazyOf(bytes.also { /* .. validate .. */ })`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,125 @@ import kotlinx.serialization.encoding.Encoder
...
+            if(first().toUByte()>127u) throw Asn1Exception("OID top-level arc can only be number 0, 1 or 2")
+        }
+        nodes?.apply {
```
iaik-jheher, line 43: same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,125 @@ import kotlinx.serialization.encoding.Encoder
...
+            if (size < 2) throw Asn1StructuralException("at least two nodes required!")
```
iaik-jheher, line 44: does this check not apply to the bytearray version?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -19,48 +24,125 @@ import kotlinx.serialization.encoding.Encoder
...
     /**
-     * @param oid in human-readable format (e.g. "1.2.96")
+     * Creates an OID in the 2.25 subtree that requires no formal registration.
+     * E.g. the UUID `550e8400-e29b-41d4-a716-446655440000` results in the OID
+     * `2.25.113059749145936325402354257176981405696`
      */
-    constructor(oid: String) : this(*(oid.split(if (oid.contains('.')) '.' else ' ')).map { it.toUInt() }.toUIntArray())
+    @OptIn(ExperimentalUuidApi::class)
+    constructor(uuid: Uuid) : this(
+        bytes = byteArrayOf((2 * 40 + 25).toUByte().toByte(), *uuid.toBigInteger().toAsn1VarInt()),
```
iaik-jheher, line 102: i suspect `(2 * 40 + 25)` is not compile time evaluated and would prefer if it were replaced accordingly

(maybe it's premature optimization?)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: nope; I want it to break right at the initialization, not some time later on

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: same same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: the first two nodes are encoded into a single byte (the first one), so you need discrete checks, dependeing on where you are coming from

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: i would hope, it is compile-time evaluated. Anyways, it is comprehensible, because that is the textbook way of encoding those first two nodes. If i just plug in a constant, then I'd need a comment to de-obfuscate the number

JesusMcCloud: also: we are already 1000% (not a typo) faster than my initial approach and 500% faster than before this PR and this single expression will not impact that gain

JesusMcCloud: (yes a typo, i checked the numbers:  >2000% faster, and >1000% faster, respectively)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
iaik-jheher: that _would_ break right at the initialization

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
iaik-jheher: same same same :)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: seems so. Still no, because i want all checks in one place to see them at a glance

iaik-jheher (approved):

PR #159, opened 2024-10-07
Add digest to interface

n0900:
I propose to add digest as a variable to the `SignatureAlgorithm` interface in order to prevent the following code snippet.
@JesusMcCloud told me to ask you about it.

Screenshot from 2024-10-07 17-59-15

### DIFF
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
@@ -6,21 +6,24 @@ enum class RSAPadding {
 }
 
 sealed interface SignatureAlgorithm {
+
+    val digest: Digest?
+
     data class HMAC(
         /** The digest to use */
-        val digest: Digest
+        override val digest: Digest
     ) : SignatureAlgorithm
 
     data class ECDSA(
         /** The digest to apply to the data, or `null` to directly process the raw data. */
-        val digest: Digest?,
+        override val digest: Digest?,
         /** Whether this algorithm specifies a particular curve to use, or `null` for any curve. */
         val requiredCurve: ECCurve?
     ) : SignatureAlgorithm
 
     data class RSA(
         /** The digest to apply to the data. */
-        val digest: Digest,
+        override val digest: Digest?,
         /** The padding to apply to the data. */
         val padding: RSAPadding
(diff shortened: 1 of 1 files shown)


iaik-jheher:
I had this discussion with @JesusMcCloud in the past; just because every kind of signature algorithm we currently support uses a Digest, that doesn't mean that every signature algorithm we might support in the future (PQC?) will use a Digest; so I was reluctant to add it to the interface.

We could maybe have a `SignatureAlgorithm.WithDigest` marker interface in between?

JesusMcCloud:
This is really tricky, because even signature that could have a digest, might not always come with one, so nullable and not having it at all is semantically different. 

> We could maybe have a `SignatureAlgorithm.WithDigest` marker interface in between?
This is probably the only sane way to go, because using it, as it is now, is just a pain and needs improvement. I don't think we can go wrong with this suggestion

iaik-jheher:
> I don't think we can go wrong with this suggestion

well, we've run into this before at various points... kotlin has no good way to express class traits, i.e., to say that "any instance of this class must have one or the other of this marker interface". [This](https://pl.kotl.in/V9om1-9KD) is not legal.

Also, there is no way to cleanly deal with multi-trait classes. [This](https://pl.kotl.in/hJKNrZJZ1) is also not legal.

So if you ever need two or more marker interfaces on the same type, it becomes a complete mess; you end up having to create new interfaces for each possible intersection you might need, and update every implementation's inheritance list.

I don't think this will be an issue for `SignatureAlgorithm` _now_ (since we only have a single trait that we want to add), but we need to realize that it's the _only_ trait we get to realistically add.

n0900:
> I had this discussion with @JesusMcCloud in the past; just because every kind of signature algorithm we currently support uses a Digest, that doesn't mean that every signature algorithm we might support in the future (PQC?) will use a Digest; so I was reluctant to add it to the interface.

But do we not circumvent this problem by making `digest` nullable? This simply states that it might have one.

JesusMcCloud:
see my previous comment. nullable and no digest at all are semantically different

n0900:
> see my previous comment. nullable and no digest at all are semantically different

True, but so is `X509SignatureAlgorithm(oid=..., isEc=False)` and `X509SignatureAlgorithm(oid)` however functionally they are the same

n0900:
Any updates on this? Should this be closed? @JesusMcCloud @iaik-jheher

PR #167, opened 2024-10-23
Better ASN.1 encoding

JesusMcCloud:
Efficient ASN.1 encoding

### DIFF
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
@@ -94,7 +94,7 @@ sealed class Asn1Element(
         write(element.tlv.tag.encodedTag);
         write(element.encodedLength);
         if (element.children != null) { // structure
-            element.children.forEach(this::writeAsn1Element)
+            element.children!!.forEach { writeAsn1Element(it) }
         } else { // primitive
             write(element.tlv.content)
         }
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
@@ -7,6 +7,9 @@ import at.asitplus.signum.indispensable.io.ByteArrayBase64Serializer
 import io.matthewnelson.encoding.base16.Base16
 import io.matthewnelson.encoding.core.Decoder.Companion.decodeToByteArray
 import io.matthewnelson.encoding.core.Encoder.Companion.encodeToString
+import kotlinx.io.Buffer
+import kotlinx.io.Sink
+import kotlinx.io.readByteArray
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.descriptors.PrimitiveKind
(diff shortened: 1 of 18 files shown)


indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -235,26 +255,25 @@ sealed class Asn1Element(
...
     ) : Comparable<Tag> {
-        private constructor(values: Triple<ULong, Int, ByteArray>) : this(values.first, values.second, values.third)
-        constructor(derEncoded: ByteArray) : this(
-            derEncoded.iterator().decodeTag().let { Triple(it.first, it.second.size, derEncoded) }
-        )
 
+
+        constructor(derEncoded: ByteArray) : this(derEncoded.iterator().decodeTag())
+
+        //workaround because we cannot return two values or assign params in a destructured manner
```
iaik-jheher, line 266: unclear what this means

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -235,26 +255,25 @@ sealed class Asn1Element(
 
     @Serializable
     @ConsistentCopyVisibility
-    data class Tag private constructor(
-        val tagValue: ULong, val encodedTagLength: Int,
+    data class Tag internal constructor(
```
iaik-jheher, line 258: why was the visibility changed?

JesusMcCloud: we now use this ctor directly in Asn1Decoding.kt

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: we need to "directly" pass the result of `decodeTag` to a constructor. which is a pair. you cannot destructure the return value directly in a function or constructor invocation, so we need another constructor to do the destructuring.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -78,10 +81,27 @@ sealed class Asn1Element(
...
+            write(element.tlv.tag.encodedTag);
+            write(element.encodedLength);
+            childElems.forEach { child -> writeAsn1Element(child) }
+        }
+
+            ?: also { //primitive
+                write(element.tlv.tag.encodedTag)
+                write(element.encodedLength)
+                write(element.tlv.content)
+            }
```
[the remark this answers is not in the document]
iaik-jheher: for the record i still think that we should revisit the entire `tlv` concept, since it seems that for any given element either "elm.children" or "elm.tlv.content" is used, which sounds like the two classes are so intrinsically linked they should maybe be a single class. At the very least the "has children" vs "has content" distinction should be a sealed-class diff.

either way that's a change for another PR.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: i would prefer something like `decodeTag(...).let { (a,b) -> Tag(a,b) }`, though I am also unsure why `decodeTag` doesn't return a `Tag` in the first place

iaik-jheher (approved):
conditional on test run

PR #164, opened 2024-10-21
Feature/ktxio

JesusMcCloud:
use Kotlinx-io for decoding to avoid unneeded byte copying

PR #152, opened 2024-10-01
Reformat shifting bytes

nodh:
Fixes #123

PR #168, opened 2024-10-24
Feature/kotlinx io

JesusMcCloud:
Proper version of #164


JesusMcCloud:
See [here](https://github.com/a-sit-plus/signum/runs/32004371930#r10s0). The issue is ULong decoding

JesusMcCloud:
@iaik-jheher if you would be so kind and check the ULong issue

JesusMcCloud:
@iaik-jheher please review

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -354,46 +425,53 @@ fun Boolean.Companion.decodeFromAsn1ContentBytes(bytes: ByteArray): Boolean {
...
     val length = decodeLength()
-    require(length < 1024 * 1024) { "Heap space" }
-    val value = ByteArray(length) {
-        require(hasNext()) { "Out of bytes to decode" }
-        nextByte()
-    }
-
-    return TLV(Asn1Element.Tag(tag.first,tag.second), value)
+    require(length.first >= 0L) { "Illegal length: $length" }
+    return (tag to length.first) to length.second + tag.encodedTagLength
```
iaik-jheher, line 444: these `to` constructions feel too clever by half and just make the code harder to read

use `Pair(...)` please

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -354,46 +425,53 @@ fun Boolean.Companion.decodeFromAsn1ContentBytes(bytes: ByteArray): Boolean {
  */
 fun String.Companion.decodeFromAsn1ContentBytes(bytes: ByteArray) = bytes.decodeToString()
 
+/**
+ * [Asn1Element.Tag] to the decoded length
+ */
+typealias TagAndLength = Pair<Asn1Element.Tag, Long>
```
iaik-jheher, line 431:
```suggestion
private typealias TagAndLength = Pair<Asn1Element.Tag, Long>
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -354,46 +425,53 @@ fun Boolean.Companion.decodeFromAsn1ContentBytes(bytes: ByteArray): Boolean {
...
+ * @return the decoded length and the number of bytes consumed
+ */
 @Throws(IllegalArgumentException::class)
-private fun ByteIterator.decodeLength() =
-    nextByte().let { firstByte ->
+private fun Source.decodeLength(): Pair<Long, Int> =
+    readByte().let { firstByte ->
         if (firstByte.isBerShortForm()) {
-            firstByte.toUByte().toInt()
+            firstByte.toUByte().toLong() to 1
```
iaik-jheher, line 455:
```suggestion
            Pair(firstByte.toUByte().toLong(), 1)
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -354,46 +425,53 @@ fun Boolean.Companion.decodeFromAsn1ContentBytes(bytes: ByteArray): Boolean {
...
         } else { // its BER long form!
             val numberOfLengthOctets = (firstByte byteMask 0x7F).toInt()
-            (0 until numberOfLengthOctets).fold(0) { acc, index ->
-                require(hasNext()) { "Can't decode length" }
-                acc + (nextByte().toUByte().toInt() shl Byte.SIZE_BITS * (numberOfLengthOctets - index - 1))
-            }
+            (0 until numberOfLengthOctets).fold(0L) { acc, index ->
+                require(!exhausted()) { "Can't decode length" }
+                acc + (readUByte().toLong() shl Byte.SIZE_BITS * (numberOfLengthOctets - index - 1))
+            } to 1 + numberOfLengthOctets
```
iaik-jheher, line 461:
```suggestion
            }.let { Pair(it, 1 + numberOfLengthOctets) }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -3,13 +3,24 @@ package at.asitplus.signum.indispensable.asn1.encoding
...
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
+internal val UVARINT_SINGLEBYTE_MAXVALUE_UBYTE: UByte = 0x80u
+internal val UVARINT_SINGLEBYTE_MAXVALUE: Byte = 0x80.toByte()
+
+internal val UVARINT_MASK = 0x7F
+internal val UVARINT_MASK_UBYTE: UByte = 0x7Fu
+internal val UVARINT_MASK_UINT: UInt = 0x7Fu
+internal val UVARINT_MASK_ULONG: ULong = 0x7FuL
```
iaik-jheher, line 22: can these not be `const`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
+/**
+ * Encodes this number using unsigned VarInt encoding as used within ASN.1:
+ * Groups of seven bits are encoded into a byte, while the highest bit indicates if more bytes are to come.
+ *
+ * This kind of encoding is used to encode [ObjectIdentifier] nodes and ASN.1 Tag values > 30
+ */
+fun UInt.toAsn1VarInt(): ByteArray = throughBuffer { it.writeAsn1VarInt(this) }
+
+private fun MutableList<Byte>.asn1VarIntByteMask(it: Int) =
+    (if (isLastIndex(it)) 0x00 else UVARINT_SINGLEBYTE_MAXVALUE).toByte()
```
iaik-jheher, line 184: is varint really aware of the underlying `MutableList` we are using to store bytes? i would expect `lastByte & 0x80` to simply be an error case (since it indicates there should be more bytes, but there are not)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
+/**
+ * Encodes this number using varint encoding as used within ASN.1: groups of seven bits are encoded into a byte,
+ * while the highest bit indicates if more bytes are to come
+ *
+ * @return the number of bytes written to the sink
+ */
+private fun Sink.writeAsn1VarInt(number: ULong, bits: Int):Int {
+    if (number < 128u) return writeByte(number.toByte()).run { 1 } //Fast case
     var offset = 0
     var result = mutableListOf<Byte>()
```
iaik-jheher, line 264: there has to be a way to refactor this and avoid the list

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
- * This kind of encoding is used to encode [ObjectIdentifier] nodes and ASN.1 Tag values > 30
+ * @return the number of bytes written to the sink
  */
-fun UInt.toAsn1VarInt(): ByteArray {
-    if (this < 128u) return byteArrayOf(this.toByte()) //Fast case
+@Throws(IllegalArgumentException::class)
+fun Sink.writeAsn1VarInt(number: BigInteger): Int {
+    if (number.isZero()) return writeByte(0).run { 1 }
+    require(number.isPositive) { "Only positive Numbers are supported" }
+    if (number < UVARINT_SINGLEBYTE_MAXVALUE) return writeByte(number.byteValue(exactRequired = true)).run { 1 } //Fast case
```
iaik-jheher, line 289:
```suggestion
    require(!number.isNegative) { "Only non-negative numbers are supported" }
    if (number < UVARINT_SINGLEBYTE_MAXVALUE) { // Fast case
        writeByte(number.byteValue(exactRequired = true))
        return 1
    }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
     var offset = 0
     var result = mutableListOf<Byte>()
```
iaik-jheher, line 291: this function can be replaced with:
```kotlin
val numBytes = (number.bitLength()+6)/7 // division rounding up
(numBytes-1).downTo(0).forEach { byteIndex ->
    writeByte(
        ((number shr (byteIndex*7)).byteValue(exactRequired = false) and UVARINT_MASK_BIGINT) or
            (if (byteIndex > 0) UVARINT_SINGLEBYTE_MAXVALUE else 0)))
}
return numBytes
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
-private fun MutableList<Byte>.fromBack(it: Int) = this[size - 1 - it]
-
-
 /**
- * Decodes an ULong from bytes using varint encoding as used within ASN.1: groups of seven bits are encoded into a byte,
- * while the highest bit indicates if more bytes are to come. Trailing bytes are ignored.
+ * Decodes an ASN.1 unsigned varint to an [ULong], copying all bytes from the source into a [ByteArray].
  *
- * @return the decoded ULong and the underlying varint-encoded bytes as `ByteArray`
+ * @return the decoded [ULong] and the underlying varint-encoded bytes as ByteString
```
iaik-jheher, line 310: comment says `ByteString`, function signature says `ByteArray`, which is it?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
+fun Source.decodeAsn1VarULong(): Pair<ULong, ByteArray> = decodeAsn1VarInt(ULong.SIZE_BITS)
 
 /**
- * Decodes an ULong from bytes using varint encoding as used within ASN.1: groups of seven bits are encoded into a byte,
- * while the highest bit indicates if more bytes are to come. Trailing bytes are ignored.
+ * Decodes an ASN.1 unsigned varint to an [UInt], copying all bytes from the source into a [ByteString].
  *
- * @return the decoded ULong and the underlying varint-encoded bytes as `ByteArray`
- * @throws IllegalArgumentException if the number is larger than [ULong.MAX_VALUE]
+ * @return the decoded [UInt] and the underlying varint-encoded bytes as ByteString
```
iaik-jheher, line 319: same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
+        accumulator.writeUByte(current)
+        if (current >= UVARINT_SINGLEBYTE_MAXVALUE_UBYTE) {
+            result = (current and UVARINT_MASK_UBYTE).toULong() or (result shl 7)
         } else {
-            result = (current and 0x7F.toUByte()).toULong() or (result shl 7)
+            result = (current and UVARINT_MASK_UBYTE).toULong() or (result shl 7)
             break
         }
-        if (++offset > ceil(ULong.SIZE_BYTES.toFloat() * 8f / 7f)) throw IllegalArgumentException("Tag number too Large do decode into ULong!")
+        if (++offset > ceil((bits * 8).toFloat() * 8f / 7f)) throw IllegalArgumentException("Tag number too Large do decode into $bits bits!")
```
iaik-jheher, line 349: this is a generic asn.1 varint decoder, the exception message should not mention tags

also `do decode` -> `to decode`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
 }
 
-
 /**
- * Decodes an unsigned BigInteger from bytes using varint encoding as used within ASN.1: groups of seven bits are encoded into a byte,
- * while the highest bit indicates if more bytes are to come. Trailing bytes are ignored.
+ * Decodes an ASN.1 unsigned varint to a [BigInteger], copying all bytes from the source into a [ByteString].
  *
- * @return the decoded unsigned BigInteger and the underlying varint-encoded bytes as `ByteArray`
+ * @return the decoded [BigInteger] and the underlying varint-encoded bytes as ByteString
```
iaik-jheher, line 358: same as previously, `ByteArray` or `ByteString`, which is it?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
  */
-inline fun Iterable<Byte>.decodeAsn1VarBigInt(): Pair<BigInteger, ByteArray> = iterator().decodeAsn1VarBigInt()
+fun Source.decodeAsn1VarBigInt():Pair<BigInteger, ByteArray> {
+    var result = BigInteger.ZERO
+    val accumulator = Buffer()
+    while (!exhausted()) {
+        val curByte = readByte()
+        val current = BigInteger(curByte.toUByte().toInt())
```
iaik-jheher, line 365: do we really need a `BigInteger` instantiation for a single byte here?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
+        writeByte(number.toByte())
+        7 + byteWritten
     }
 
-    return result to accumulator.toByteArray()
 }
 
+private inline fun Sink.writeOrSkipPadding(number: Long, shift: Int, padded: Boolean): Int {
+    val byte = (number ushr shift).toByte()
+    val writeFirstByte = padded || (byte != 0.toByte())
```
iaik-jheher, line 462: why is this an `or`? should it not be `and`? as i understand the `padded` param to writevarint, it can be used to _suppress_ the leading zero (the leading zero should _only_ be written if it is both _needed_ and _padded = true_.)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -241,222 +163,374 @@ fun Int.toUnsignedByteArray() = toLong().toUnsignedByteArray()
...
-        result = (current and mask) or (result shl 7)
-        if (current < 0x80.toUByte()) break
+private fun Sink.writeTwosComplement(number: Long, padded: Boolean = true): Int = when {
+    (number >= -0x80L && number <= 0x7FL) -> {
+        writeByte(number.toByte())
+        1
+    }
+
+    (number >= -0x8000L && number <= 0x7FFFL) -> {
+        val byteWritten = writeOrSkipPadding(number, 8, padded)
```
iaik-jheher, line 396: i do not see how this can ever write a leading zero byte. (also: please do not refer to a leading zero byte as "padding"; it is not padding.)

the matched `when` condition ensures that the number fits in 2 bytes without leading zero; leading zero is needed for range `0x8000` to `0xffff`

also why are we not writing the second byte? this entire function feels wrong the more i look at it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
iaik-jheher: i also think that this is a hack, and would suggest either introducing a `writeUnsignedBytes` helper, or a lightweight `SkipLeadingZeroSinkWrapper` that does nothing except omit the first write if it is zero.

both approaches are far cleaner than this mess

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +129,21 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
-@Suppress("NOTHING_TO_INLINE")
-inline fun ByteArray.asBuffer()= Buffer().also{it.write(this)}
-
 @OptIn(UnsafeIoApi::class)
 internal fun ByteArray.wrapInUnsafeSource(): Source = Buffer().apply {
     UnsafeBufferOperations.moveToTail(this, this@wrapInUnsafeSource)
 }
+
+@OptIn(UnsafeIoApi::class)
+internal fun wrapInUnsafeSource(vararg bytes: ByteArray) = Buffer().apply {
```
iaik-jheher, line 138: add a kdoc describing why this is unsafe

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +129,21 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
+}
+
+/**
+ * Helper to create a buffer, operate on it and return its contents as a [ByteArray]
+ */
+internal inline fun throughBuffer(operation: (Buffer) -> Unit): ByteArray =
+    Buffer().also { operation(it) }.readByteArray()
+
+internal inline fun <reified T> ByteArray.throughBuffer(operation: (Source) -> T): T =
+    wrapInUnsafeSource().let { operation(it) }
```
iaik-jheher, line 149:
```suggestion
    wrapInUnsafeSource().let(operation)
```

iaik-jheher (changes requested):

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## 3.0
 
-### 3.10.0 NEXT (Supreme 0.5.0 NEXT)
+### 3.10.0 NEXT (Supreme 0.5.0 NEXT) Breaking Changes Ahead!
```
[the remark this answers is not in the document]
JesusMcCloud: bouncy castle manages this with every other minor release. Still i'll rephrase!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -491,6 +467,40 @@ sealed class Asn1Structure(
...
+    }
+
+    override fun prettyPrintContents(indent: Int): String =
+        children.joinToString(
+            prefix = "\n" + (" " * indent) + "{\n",
+            separator = "\n",
+            postfix = "\n" + (" " * indent) + "}"
+        ) { it.prettyPrint(indent + 2) }
+
+    override fun contentToString(): String = (if (isSorted) " SORTED " else "") + ", children=$children"
```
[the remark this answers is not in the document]
JesusMcCloud: you cannnot assume compliant structures in the real world, so this must stay, otherwise we'll mess it up when re-encoding

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -491,6 +467,40 @@ sealed class Asn1Structure(
...
+
+    override fun hashCode() = 31 * super.hashCode() + children.hashCode()
+
+    /**
+     * the [isSorted] flag has no bearing on equals!
```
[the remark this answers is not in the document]
JesusMcCloud: the real world is a nasty place to be

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -58,40 +59,39 @@ class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
      * for details.
      * Lazily evaluated.
      */
-    val bytes: ByteArray by if (bytes != null) lazyOf(bytes) else lazy {
-        this.nodes.toOidBytes()
-    }
+    val bytes: ByteArray by lazy { bytes ?: this.nodes.toOidBytes() }
```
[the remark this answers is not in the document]
JesusMcCloud: i know. the old version was inefficient, because it initialized eagerly. why would we want this? are we missing early sanity checks if we init lazily?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -10,91 +10,162 @@ import at.asitplus.signum.indispensable.asn1.BERTags.T61_STRING
...
 @Throws(Asn1Exception::class)
-fun Asn1Element.Companion.parse(input: ByteIterator): Asn1Element = parseFirst(input).also {
-    if (input.hasNext()) throw Asn1StructuralException("Trailing bytes found after the first ASN.1 element")
-}
+@Deprecated(
+    "Use a ByteArray or (even better) a kotlinx.io Source as input when possible. This method copies all bytes from the input twice and is inefficient.",
+    ReplaceWith("source.readAsn1Element(); require(source.exhausted())")
+)
+fun Asn1Element.Companion.parse(input: ByteIterator): Asn1Element =
+    parse(mutableListOf<Byte>().also { while (input.hasNext()) it.add(input.nextByte()) }.toByteArray())
```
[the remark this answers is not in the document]
JesusMcCloud: i don't think there is

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -10,91 +10,162 @@ import at.asitplus.signum.indispensable.asn1.BERTags.T61_STRING
...
+ *
  */
+@Deprecated(
+    "Use a ByteArray or (even better) a kotlinx.io Source as input when possible. This method copies all bytes from the input twice and is inefficient.",
+    ReplaceWith("source.readFullyToAsn1Elements()"),
+    DeprecationLevel.WARNING
+)
 @Throws(Asn1Exception::class)
-fun Asn1Element.Companion.parseAll(input: ByteIterator): List<Asn1Element> = input.doParseAll()
+fun Asn1Element.Companion.parseAll(input: ByteIterator): List<Asn1Element> =
```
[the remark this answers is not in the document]
JesusMcCloud: same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: be my guesst

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -10,91 +10,162 @@ import at.asitplus.signum.indispensable.asn1.BERTags.T61_STRING
...
-//this only makes sense until we switch to kotlinx.io
 @Throws(Asn1Exception::class)
-fun Asn1Element.Companion.parseFirst(input: ByteIterator): Asn1Element = input.doParseSingle()
+fun Source.readFullyToAsn1Elements(): Pair<List<Asn1Element>, Long> = mutableListOf<Asn1Element>().let { list ->
+    var bytesRead = 0L
+    while (!exhausted()) readAsn1Element().also { (elem, nBytes) ->
+        bytesRead += nBytes
+        list.add(elem)
+    }
+    list to bytesRead
```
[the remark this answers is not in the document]
JesusMcCloud: if you insinst

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
JesusMcCloud: you really maintain `Pair((tag to length.first), (length.second + tag.encodedTagLength)` is more readable?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
JesusMcCloud: i don't care one way or the other

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
JesusMcCloud: i don't care one way or the other

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: it is now a non-mutable list and unless i am mistaken or I messed something up, you approved this logic [here](https://github.com/a-sit-plus/signum/pull/152).

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -24,25 +26,18 @@ import kotlin.native.ObjCName
  */
 @Serializable(with = Asn1EncodableSerializer::class)
 sealed class Asn1Element(
-    internal val tlv: TLV,
-    protected open val children: List<Asn1Element>?
+    val tag: Tag
 ) {
 
     override fun equals(other: Any?): Boolean {
```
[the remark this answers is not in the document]
iaik-jheher: i do not believe that this override does anything that is not provided by base class `equals`? i could be wrong

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -68,85 +63,66 @@ sealed class Asn1Element(
...
     /**
      * Total number of bytes required to represent the ths element, when encoding to ASN.1.
      */
-    val overallLength by lazy { length + tlv.tag.encodedTagLength + encodedLength.size }
+    val overallLength by lazy { length + tag.encodedTagLength + encodedLength.size }
 
-    protected open val content by lazy { tlv.content }
 
-    val tag by lazy { tlv.tag }
+    protected val derEncodedLazy = lazy { Buffer().also { encodeTo(it) }.readByteArray() }
```
iaik-jheher, line 74: could be `private` i believe

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -592,16 +602,25 @@ class Asn1CustomStructure private constructor(
     ) : this(children, tag.toULong(), tagClass, sortChildren)
 
 
-    override val content: ByteArray by lazy {
+    /**
+     * Raw byte DER-encoded representation of this custom structure's children.
+     * This property is `null` **unless** the `CONSTRUCTED` flag of this structure's tag is overridden to `false`
+     */
+    val content: ByteArray? by lazy {
```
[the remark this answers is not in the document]
iaik-jheher: could we just use `this.derEncoded` there? either way i am not too bothered since `toString` is debug case, not production case

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
iaik-jheher: the class is initialized with either `bytes` or `nodes` set. why would we capture the constructor parameter in a lambda, then wait for the first access to actually return it, if we already know at construction time what the value will be?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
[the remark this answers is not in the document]
iaik-jheher: write a two-line wrapper for `ByteIterator` -> `RawSource` then? it's better than copying everything

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
iaik-jheher: `Pair(TagAndLength(tag, length.first), length.second + tag.encodedTagLength)`, but yes

iaik-jheher: PS if you think it's too much, then split it up into `val tagInfo = ...`, `val parsedLength = ...`, and `Pair(tagInfo, parsedLength)`... i agree it's a mouthful, but it's still a mouthful if you use `to`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
iaik-jheher: already covered by the `if (number < UVARINT_SINGLEBYTE_MAXVALUE)` branch above (assuming you kept it)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: ld;dr: yes!

you will need a bigint three lines down anyways, because you cant `or` a bigint with a non-bigint

iaik-jheher: are you sure that `bigint or long` also doesn't work? that would be weird

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: how would you craft such a wrapper, given we write each byte individually?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: you are wrong

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
iaik-jheher: not sure what the sink interface looks like offhand, but
```
class SinkWrapper(val wrapped: Sink) {
    var isFirst = True
    override fun writeByte(it: Byte) {
        val skip = (isFirst && (it == 0))
        isFirst = false;
        if(!skip) wrapped.writeByte(it)
    }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
iaik-jheher: what does it provide? i'm confused

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: i don't want remove the `val`, because i think it brings added value to the content string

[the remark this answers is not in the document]
JesusMcCloud: and it also brings added value dealing with custom structures, because you can check if they adhere to SET or SEQUENCE semantics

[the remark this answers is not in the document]
iaik-jheher: but in my understanding, `isSorted` does not actually convey whether the thing is sorted, but only whether _we_ sorted it, right? so a pre-sorted (compliant) set would still not show "SORTED"?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: reworked as per our offline discussion

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: reworked as per our offline discussion

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
JesusMcCloud: reworked as per out offline discussion

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
JesusMcCloud: because maybe we will never access the bytes, but just operate on the nodes, if we require it for some intermediate FOO without actually encoding it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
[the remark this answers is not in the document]
JesusMcCloud: same difference, because the first thing that will happen is, you make a call to `.buffered()` to make the source usable and then it tries to copy 8KiB into an internal buffer

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
JesusMcCloud: i don't care enough ;-)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: Unless overlooked something due to borked Github rendering of large PRs with many comments, this is the only outstanding issue and i am not the right person to wrap their head around this

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
iaik-jheher: but when you operate on the nodes, you lazy-init the nodes by looking at `this.bytes` (which lazy-inits `this.bytes` to the predetermined non-null ctor parameter)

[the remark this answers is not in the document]
iaik-jheher: so the idea is that you always directly store whichever parameter is non-null, and the other parameter is lazy-initialized based on the non-null parameter

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: true, yes. maybe we need to rename

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
JesusMcCloud: default equals does nothing meaningful unless it is a data class

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
JesusMcCloud: be my guest to double-check the sources of bignum. I did not see it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -321,142 +279,238 @@ private fun MutableList<Byte>.fromBack(it: Int) = this[size - 1 - it]
...
+ * integer representation of a UUID, chances are high, though, that it works with random positive BigIntegers between
+ * 16 and 14 bytes large.
  *
- * @return the decoded unsigned BigInteger and the underlying varint-encoded bytes as `ByteArray`
+ * Returns `null` if conversion fails. Never throws.
  */
-inline fun Iterable<Byte>.decodeAsn1VarBigInt(): Pair<BigInteger, ByteArray> = iterator().decodeAsn1VarBigInt()
+@OptIn(ExperimentalUuidApi::class)
+fun Uuid.Companion.fromBigintOrNull(bigInteger: BigInteger): Uuid? =
+    catching { fromByteArray(bigInteger.toByteArray().ensureSize(16)) }.getOrNull()
```
iaik-jheher, line 318:
```suggestion
    catchingUnwrapped { fromByteArray(bigInteger.toByteArray().ensureSize(16)) }.getOrNull()
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
iaik-jheher: copy from `Sink.writeAsn1VarInt(number: BigInteger)`, with `bitLength` replacement as discussed

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -40,14 +38,17 @@ val Base64Strict = Base64(config = Base64ConfigBuilder().apply {
 
 sealed class TemplateSerializer<T>(serialName: String = "") : KSerializer<T> {
     protected val realSerialName =
-        serialName.ifEmpty { this::class.simpleName
-            ?: throw IllegalArgumentException("Anonymous classes must specify a serialName explicitly") }
+        serialName.ifEmpty {
```
iaik-jheher, line 41: please revert whitespace-only changes to this file

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +137,43 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
-
 @OptIn(UnsafeIoApi::class)
 internal fun ByteArray.wrapInUnsafeSource(): Source = Buffer().apply {
     UnsafeBufferOperations.moveToTail(this, this@wrapInUnsafeSource)
 }
+
+/**
+ * Directly moves the byte array to a buffer without copying. Thus, it keeps bytes managed by a Buffer accessible.
+ * The bytes may be overwritten through the Buffer or even recycled to be used by another buffer.
+ * Also operating on these bytes after wrapping leads to undefined behaviour.
```
iaik-jheher, line 148:
```suggestion
 * Therefore, operating on these bytes after wrapping leads to undefined behaviour.
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +137,43 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
+ * [startIndex] is inclusive, [endIndex] is exclusive.
+ */
+@OptIn(UnsafeIoApi::class)
+internal fun wrapInUnsafeSource(bytes: ByteArray, startIndex: Int = 0, endIndex: Int = bytes.size) = Buffer().apply {
+    require(startIndex in 0..endIndex) { "StartIndex bust be between 0 and $endIndex" }
```
iaik-jheher, line 153:
```suggestion
    require(startIndex in 0..<endIndex) { "StartIndex bust be between 0 and $endIndex" }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +137,43 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
+
+
+/**
+ * Directly appends [bytes] to this Sink's internal Buffer without copying. Thus, it keeps bytes managed by a Buffer accessible.
+ * The bytes may be overwritten through the Buffer or even recycled to be used by another buffer.
+ * Also operating on these bytes after wrapping leads to undefined behaviour.
+ * [startIndex] is inclusive, [endIndex] is exclusive.
+ */
+internal fun Sink.appendUnsafe(bytes: ByteArray, startIndex: Int = 0, endIndex: Int = bytes.size): Int {
+    require(startIndex in 0..endIndex) { "StartIndex bust be between 0 and $endIndex" }
```
iaik-jheher, line 174:
```suggestion
    require(startIndex in 0..<endIndex) { "StartIndex must be between 0 and $endIndex" }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -128,10 +137,43 @@ fun ByteArray.ensureSize(size: Int): ByteArray = (this.size - size).let { toDrop
...
+    Buffer().also(operation).readByteArray()
+
+internal inline fun <reified T> ByteArray.throughBuffer(operation: (Source) -> T): T =
+    wrapInUnsafeSource().let { operation(it) }
+
+
+/**
+ * Directly appends [bytes] to this Sink's internal Buffer without copying. Thus, it keeps bytes managed by a Buffer accessible.
+ * The bytes may be overwritten through the Buffer or even recycled to be used by another buffer.
+ * Also operating on these bytes after wrapping leads to undefined behaviour.
```
iaik-jheher, line 170:
```suggestion
 * The bytes may be overwritten through the Buffer, or even recycled to be used by another buffer.
 * Thus, operating on these bytes after wrapping leads to undefined behaviour.
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
iaik-jheher: is this working now?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -10,99 +10,171 @@ import at.asitplus.signum.indispensable.asn1.BERTags.T61_STRING
...
+        bytesRead += nBytes
+        list.add(elem)
+    }
+    Pair(list, bytesRead)
+}
 
+/**
+ * Reads a [TagAndLength] and the number of consumed bytes from the source without consuming it
+ */
+fun Source.peekTagAndLen() = peek().readTagAndLength()
```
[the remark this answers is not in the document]
JesusMcCloud: almost certainly yes

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -10,99 +10,171 @@ import at.asitplus.signum.indispensable.asn1.BERTags.T61_STRING
...
+        val peekTagAndLen = peekTagAndLen()
+        val numberOfNextBytesRead = peekTagAndLen.second + peekTagAndLen.first.length
+        if (nBytesRead + numberOfNextBytesRead > nBytes) break
+        skip(peekTagAndLen.second.toLong()) // we only peeked before, so now we need to skip,
+        //                                     since we want to recycle the result below
+        val (elem, read) = readAsn1Element(peekTagAndLen.first, peekTagAndLen.second)
+        list.add(elem)
+        nBytesRead += read
+    }
+    require(nBytesRead == nBytes) { (if (nBytesRead < nBytes) "Less " else "More ") + " bytes read ($nBytesRead) than allowed ($nBytes)" }
```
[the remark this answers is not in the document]
JesusMcCloud: also says "fewer" now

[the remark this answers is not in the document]
iaik-jheher: should maybe just say `indicated length ($nBytes) does not correspond to an asn.1 element boundary`, since i think that's what it's functionally used for?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/NumberEncoding.kt
```diff
@@ -343,23 +344,40 @@ fun Sink.writeAsn1VarInt(number: UInt) = writeAsn1VarInt(number.toULong(), UInt.
...
-        writeByte(result.fromBack(index) or result.asn1VarIntByteMask(index))
+    val numBytes = (number.bitLength + 6) / 7 // division rounding up
+    (numBytes - 1).downTo(0).forEach { byteIndex ->
+        writeUByte(
+            ((number shr (byteIndex * 7)).toUByte() and UVARINT_MASK_UBYTE) or
+                    (if (byteIndex > 0) UVARINT_SINGLEBYTE_MAXVALUE_UBYTE else 0u)
+        )
     }
-    return result.size
+    //otherwise we won't ever write zero
```
iaik-jheher, line 354: this should probably just be a `if (number == 0)` at the top

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
iaik-jheher: this is still unaddressed

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 ### 3.10.0 NEXT (Supreme 0.5.0 NEXT)
```
iaik-jheher, line 6: i believe you deleted a paragraph here by accident in 6b6460a

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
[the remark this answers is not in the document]
iaik-jheher: this is unaddressed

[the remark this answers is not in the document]
JesusMcCloud: what do you propose? renaming the parameter? adapting the toString, so it is clearer? both?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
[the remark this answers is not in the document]
JesusMcCloud: then maybe i misunderstood. what is there to address?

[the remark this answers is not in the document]
iaik-jheher: the change should be reverted; the field should be `lazy` initialized only if the constructor parameter is non-null

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -433,11 +440,24 @@ sealed class Asn1Structure(
...
+     */
+    val shouldBeSorted: Boolean
 ) :
     Asn1Element(tag) {
 
-    val children: List<Asn1Element> = if (!isSorted) children else children.sortedBy { it.tag }
+    val children: List<Asn1Element> = if (!sortChildren) children else children.sortedBy { it.tag }
+
+    /**
+     * indicated whether the structure's children are aurally sorted.
```
iaik-jheher, line 457: image


i think you mean "actually"

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -482,12 +502,13 @@ sealed class Asn1Structure(
             postfix = "\n" + (" " * indent) + "}"
         ) { it.prettyPrint(indent + 2) }
 
-    override fun contentToString(): String = (if (isSorted) " SORTED " else "") + ", children=$children"
+    override fun contentToString(): String =
+        (if (shouldBeSorted) " should be SORTED " else "") + (if (isActuallySorted) " (is acutely SORTED) " else " (is NOT SORTED) ") + ", children=$children"
```
iaik-jheher, line 506: image

i assume this should also be "actually"

iaik-jheher: i would also shorten this; maybe
```
when {
    shouldBeSorted && isActuallySorted -> "SORTED"
    shouldBeSorted && !isActuallySorted -> "NON-COMPLIANT (UNSORTED)"
    else -> ""
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -482,12 +502,13 @@ sealed class Asn1Structure(
...
 
     override fun hashCode() = 31 * super.hashCode() + children.hashCode()
 
     /**
-     * the [isSorted] flag has no bearing on equals!
+     * the [doSort] flag has no bearing on equals!
```
iaik-jheher, line 511: shouldBeSorted?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -567,36 +593,42 @@ class Asn1Sequence internal constructor(children: List<Asn1Element>) :
...
 ) :
-    Asn1Structure(tag, children, sortChildren) {
+    Asn1Structure(tag, children, sortChildren, shouldBeSorted) {
     /**
      * ASN.1 CONSTRUCTED with custom tag
      * @param children the elements to put into this sequence
      * @param tag the custom tag to use
      * @param tagClass the tag class to use for this custom tag. defaults to [TagClass.UNIVERSAL]
      * @param sortChildren whether to sort the passed child nodes. defaults to false
+     * @param shouldBeSorted whether tge child nodes of this structure should be sorted according to this structure's definition.
```
iaik-jheher, line 605: tge -> the

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: i do love me some autocorrect!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -567,36 +593,42 @@ class Asn1Sequence internal constructor(children: List<Asn1Element>) :
...
+        shouldBeSorted: Boolean = false
+    ) : this(Tag(tag, constructed = true, tagClass), children, sortChildren, shouldBeSorted)
 
     /**
      * ASN.1 CONSTRUCTED with custom tag
      * @param children the elements to put into this sequence
      * @param tag the custom tag to use
      * @param tagClass the tag class to use for this custom tag. defaults to [TagClass.UNIVERSAL]
      * @param sortChildren whether to sort the passed child nodes. defaults to false
+     * @param shouldBeSorted whether tge child nodes of this structure should be sorted according to this structure's definition.
```
iaik-jheher, line 622: tge -> the

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: is does sort it if you set it to true…

JesusMcCloud: nvm. it is more consitent that way

iaik-jheher: clarification: these are different flags; `shouldBeSorted` is the `val`, `doSort` is the constructor parameter that indicates whether we are parsing

iaik-jheher (approved):

iaik-jheher (approved):

CHANGELOG.md
```diff
@@ -13,6 +19,11 @@
 * Base OIDs on BigInteger instead of UInt
 * Directly support UUID-based OID creation
 * Implement hash-to-curve and hash-to-scalar as per RFC9380
+* Use kotlinx-io as primary source for parsing
+    * Base number encoding/decoding on koltinx-io
```
nodh, line 23: typo

CHANGELOG.md
```diff
@@ -300,7 +311,7 @@ the Tag class just cannot be directly accessed from Swift and ObjC any more.
 * Proper BIT STRING
 * BitSet (100% Kotlin BitSet implementation)
 * Recursively parsing (and encapsulating) ASN.1 structures in OCTET Strings
-* Initial pretty-printing of ASN.1 Structures
+* Initial pretty-printing of ASN.1 Strucutres
```
nodh, line 314: typo

JesusMcCloud: thanks! fixed in another pr

PR #169, opened 2024-10-24
Secure hashing to curve per RFC9380

iaik-jheher:
[RFC](https://www.rfc-editor.org/rfc/rfc9380). This is needed for lots of advanced cryptography, including OPRFs and OPAQUE. Probably a good thing to have in one place.

Note: side channel security relies on the underlying BigInteger implementation being constant time. I have my doubts whether our underlying BigInteger is constant time.


JesusMcCloud:
I hate to overcomplicate things, but I really think we need some additional interface on an elliptic curve class to denote which ones work for our current EC math and which ones don't. Currently, all curves work, but this will change in the future. If we add this now, its harder to shoot your future self in the foot

iaik-jheher:
yes, i've thought about this as well; would mean refactoring the `ECCurve` enum as a sealed interface tree of objects

iaik-jheher:
> yes, i've thought about this as well; would mean refactoring the `ECCurve` enum as a sealed interface tree of objects

i don't think this is relevant to this PR though; it would be a separate refactor
(we will need this anyway when we add pairing curves etc.)

supreme/src/commonMain/kotlin/at/asitplus/signum/ecmath/RFC9380.kt
```diff
@@ -0,0 +1,219 @@
...
+
+private typealias expand_message = (msg: Sequence<ByteArray>, domain: ByteArray, lenInBytes: Int) -> Buffer
+
+private infix fun ByteArray.xor(other: ByteArray): ByteArray {
+    check(this.size == other.size)
+    return ByteArray(this.size) { i -> this[i] xor other[i] }
+}
+
+private inline fun I2OSP(value: Int, len: Int): ByteArray {
+    check(len == 2)
```
JesusMcCloud, line 54: why is this a param, when it has to be a constant value?

supreme/src/commonMain/kotlin/at/asitplus/signum/ecmath/RFC9380.kt
```diff
@@ -0,0 +1,219 @@
...
+private inline fun expand_message_xmd(digest: Digest): (msg: Sequence<ByteArray>, domain: ByteArray, lenInBytes: Int) -> Buffer {
+    check(digest.outputLength.bitSpacing == 0u) { "RFC9380 requirement: b mod 8 = 0 " }
+    check(digest.outputLength <= digest.inputBlockSize) { "RFC9380 requirement: b <= s" }
+    val sInBytes = digest.inputBlockSize.bytes.toInt()
+    val bInBytes = digest.outputLength.bytes.toInt()
+    return { msg, domain, lenInBytes ->
+        val ell = (lenInBytes.floorDiv(bInBytes) + if (lenInBytes.mod(bInBytes) != 0) 1 else 0).also {
+            require (it <= 255) { "RFC 9380 requirement: ell <= 255"}
+        }.toUByte()
+        require(lenInBytes <= 65535 && domain.size <= 255) { "RFC 9380 requirements" }
```
JesusMcCloud, line 68: the others explain the violated constraint, this one doesn't

JesusMcCloud (changes requested):
i know the snake_case function names come from the rfc, but still. please use CamelCase (but feel free to add the snake_case names into the apidocs. (yes, even for private functions)

supreme/src/commonMain/kotlin/at/asitplus/signum/ecmath/RFC9380.kt
iaik-jheher: because it has a param in the respective RFC; I2OSP is fixed-length integer encoding for context confusion avoidance
i just only implemented it for len == 2 because that's all we need

iaik-jheher: could call it `I2OSP_2` i guess

supreme/src/commonMain/kotlin/at/asitplus/signum/ecmath/RFC9380.kt
JesusMcCloud: it should spell out the requirement, just as everywhere else

iaik-jheher: done

supreme/src/commonMain/kotlin/at/asitplus/signum/ecmath/RFC9380.kt
iaik-jheher: done

iaik-jheher:
> i know the snake_case function names come from the rfc, but still. please use CamelCase (but feel free to add the snake_case names into the apidocs. (yes, even for private functions)

i really don't like this; it makes it harder to verify the code by just putting it side-by-side with the RFC. it's a line-by-line translation wherever possible on purpose.

iaik-jheher:
made the requested changes

JesusMcCloud (approved):

JesusMcCloud (approved):
🙂

PR #175, opened 2024-10-28
ECCurve refactor draft

iaik-jheher:
some early work on the ECCurve refactor discussed in #169


JesusMcCloud:
#266 is also relevant

PR #178, opened 2024-10-29
All platform signers are attestable

### LINKED ISSUE #177: fix attestable signer needing cast
Quick fix if you do it, thrice as long if I do it

iaik-jheher:
update type hierarchy to reflect this. closes #177.


JesusMcCloud (changes requested):
A simple test, illustrating this would be nice

iaik-jheher:
what are you thinking of? this is a change in type hierarchy, how should i test this? the test would trivially always succeed

JesusMcCloud (approved):

PR #193, opened 2024-11-12
Cose Signature Input

### LINKED ISSUE #147: Streamline and align Cose Signature Creation + Verification
JwsSigned is working out really smoothly, but CoseSigned, leaves much to be desired. Looking into VC-K constructing the signature input (also during verification) is rather fumbly, compared to JwsSigned.

@nodh assigning to you, since you know more about COSE than me.

nodh:
Fixes #147

### DIFF
--- a/docs/docs/examples.md
+++ b/docs/docs/examples.md
@@ -69,29 +69,28 @@ val protectedHeader = CoseHeader(
 val payload = byteArrayOf(0xC, 0xA, 0xF, 0xE)
 ```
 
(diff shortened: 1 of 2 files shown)


JesusMcCloud:
Do we want more flexibility (protected, vs. unprotected headers)? I seem to recall that we handle unprotected headers too in VC-K. If I'm right, I think we should offer both variants through shorthands

nodh:
The unprotected headers are not included in the signature input since they're ... well ... not protected by the signature.

JesusMcCloud (approved):

PR #194, opened 2024-11-13
COSE Signed API

nodh:
Adds a type parameter to `CoseSigned`. While not as thorough as for `JwsSigned`, it's still worth it to provide some form of guidance to clients as which payload to expect.

See also https://github.com/a-sit-plus/vck/pull/165

### DIFF
--- a/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
+++ b/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
@@ -5,6 +5,7 @@ import at.asitplus.catching
 import at.asitplus.signum.indispensable.*
 import at.asitplus.signum.indispensable.cosef.io.Base16Strict
 import at.asitplus.signum.indispensable.cosef.io.ByteStringWrapper
+import at.asitplus.signum.indispensable.cosef.io.ByteStringWrapperSerializer
 import at.asitplus.signum.indispensable.cosef.io.coseCompliantSerializer
 import at.asitplus.signum.indispensable.pki.X509Certificate
 import io.matthewnelson.encoding.core.Encoder.Companion.encodeToString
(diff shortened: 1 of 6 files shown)


indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
```diff
@@ -1,83 +1,78 @@
...
     val signature: CryptoSignature by lazy {
         if (protectedHeader.value.usesEC() ?: unprotectedHeader?.usesEC() ?: (rawSignature.size < 2048))
             CryptoSignature.EC.fromRawBytes(rawSignature)
         else CryptoSignature.RSAorHMAC(rawSignature)
     }
 
-    fun serialize() = coseCompliantSerializer.encodeToByteArray(this)
+    fun serialize(): ByteArray = coseCompliantSerializer.encodeToByteArray(CoseSignedSerializer(), this)
+
+    fun getTypedPayload(deserializationStrategy: DeserializationStrategy<P>): KmmResult<P?> = catching {
```
JesusMcCloud, line 53: This should be lazy and a test that is not already based on a bytearray would be nice

JesusMcCloud (changes requested):
Why does this behave differently from JwsSigned, that has the typed payload as property? should not the seriializer handle conversion from/to bytestringwrapper?

JesusMcCloud:
> Why does this behave differently from JwsSigned, that has the typed payload as property? should not the seriializer handle conversion from/to bytestringwrapper?

In VCK, there's bytestringwrapper all over the place, i think we can do away with it and add this wrapping logic into the serializer, while we're at it

JesusMcCloud:
@n0900 can this be made more convenient: as in: directly pass the typed payload to the constructor and have the serializer handle the byte string wrapping? I know it's different from your serialization foo, because it can by any type, but still…

nodh:
Best I can do with my knowledge of kotlinx serialization ... `CoseSignedSerializer` can't handle the payload directly, because it doesn't know about the type of the payload, since it's referenced by annotation on `CoseSigned`. In contrast, `JwsSigned` itself is not annotated with `@Serializable`, so it is easier there. I'm happy for any suggestions on how to solve the shortcomings.

JesusMcCloud (approved):

PR #199, opened 2024-11-14
Port `CryptoPublicKey.RSA` to `ASN1Integer`

iaik-jheher:


indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Integer.kt
```diff
@@ -63,6 +63,8 @@ sealed class Asn1Integer(internal val uint: VarUInt, private val sign: Sign) {
         override fun twosComplement(): ByteArray = uint.bytes.let {
             if (it.first().countLeadingZeroBits() == 0) listOf(0.toUByte()) + it else it
         }.toUByteArray().toByteArray()
+        fun bitLength(): UInt = uint.bitLength().toUInt()
+        val magnitude = uint.bytes.toUByteArray().toByteArray()
```
JesusMcCloud, line 67: please add empty lines inbetween

JesusMcCloud (changes requested):
Sure looks very tidy, but apparently it is not entirely correct.  
Also: Changelog is missing

iaik-jheher:
splitting off the asn1 int changes/fixes into #200

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -165,52 +165,34 @@ sealed class CryptoPublicKey : Asn1Encodable<Asn1Sequence>, Identifiable {
...
-            params.e
-        )
+        val bits = n.bitLength().let { Size.of(it) ?: throw IllegalArgumentException("Unsupported key size $it bits") }
 
         /**
          * @throws IllegalArgumentException in case of illegal input (odd key size, for example)
          */
         @Throws(IllegalArgumentException::class)
-        constructor(n: ByteArray, e: Int) : this(sanitizeRsaInputs(n, e))
+        constructor(n: ByteArray, e: ByteArray) : this(Asn1Integer.fromUnsignedByteArray(n), Asn1Integer.fromUnsignedByteArray(e))
```
JesusMcCloud, line 186: Since we're now using ASN.1 integer, we should document that the byte array contains only the magnitude

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -165,52 +165,34 @@ sealed class CryptoPublicKey : Asn1Encodable<Asn1Sequence>, Identifiable {
...
+
+        constructor(n: ByteArray, e: Int): this(Asn1Integer.fromUnsignedByteArray(n), Asn1Integer(e) as Asn1Integer.Positive)
```
JesusMcCloud, line 188: Since we're now using ASN.1 integer, we should document that the byte array contains only the magnitude

JesusMcCloud (approved):
Pending two lines of API docs as per my other comments

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
iaik-jheher: i removed this constructor again (it was added in this PR) since it's not used anywhere, in favor of more obvious constructors involving BigInt

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
iaik-jheher: this is just a backwards compatibility constructor at this point; i added a deprecation notice

PR #201, opened 2024-11-18
Refactor `Asn1Integer` to use `ByteArray` internally

iaik-jheher:


JesusMcCloud:
what's the status on this one?

JesusMcCloud:
ping

iaik-jheher:
good question, i honestly don't remember, i will need to review it myself

JesusMcCloud:
IIRC you would be done, were it not for some missing cleanups and now bogus tests. Just double-check whether OID equality still works after your changes. Please, review, as I would really like your changes in the mainline tree!

iaik-jheher:
I remember now; I wanted to additionally refactor it to use `ULongArray` internally since I suspect it would be a very significant speed-up.

Either way, I'm happy to have this merged as-is for now, with an additional issue tracking the fact that we should use `ULongArray`.

JesusMcCloud (approved):
I took the liberty to complete a nonsensical testcase and add equality tests.
I also created #239

PR #208, opened 2024-11-25
Small Identifier refactoring

nodh:
Is based on current state of `signum` in `vck`, and because of that the Changelog is not right.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,23 @@
 
 ## 3.0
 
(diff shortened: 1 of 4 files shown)


JesusMcCloud (approved):
Add a line to the changelog in your last commit and then cherry-pick onto development

nodh:
Merged in https://github.com/a-sit-plus/signum/commit/2ae952abc6cf3a9dda1d21ff0bb405558618c408

PR #212, opened 2024-11-29
Fix/jws signed deserialize

n0900:
iOS cannot deal with the previous kotlinx.serialization call in case this the generic P is a Polymorphic type (for instance `RequestPararameters`). 

The signature orients itself on the kotlinx.serialization signature.


n0900:
Related vck pull request https://github.com/a-sit-plus/vck/pull/173

JesusMcCloud (approved):
if it works, it works!

nodh (approved):

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsSigned.kt
```diff
@@ -85,15 +86,20 @@ data class JwsSigned<out P : Any>(
          * Deserializes the input, expected to contain a valid JWS (three Base64-URL strings joined by `.`),
          * into a [JwsSigned] with [P] as the type of the payload.
          */
-        inline fun <reified P : Any> deserialize(it: String, json: Json = Json): KmmResult<JwsSigned<P>> =
+        inline fun <reified P : Any> deserialize(
```
nodh, line 89: Does this still has to be an `inline` function?

n0900:
> isn't this a major breaking change? you are removing `deserialize<T>`, which is part of the public api?

Yes that is correct. We can make this parameter nullable but this will presumably lead to user errors which are hard to diagnose. Since the generic typing in JwsSigned is a recent change so now is probably the last good time to change it.

Your call though

iaik-jheher:
What's the reason to remove this functionality for _all_ classes if it's only broken (how?) for _polymorphic_ classes?

n0900:
> What's the reason to remove this functionality for _all_ classes if it's only broken (how?) for _polymorphic_ classes?

It does not remove functionality. For normal classes the additional parameter would be `Foo.serializer`. Again the deserialization strategy parameter could be made nullable.
As to why @JesusMcCloud probably knows; but limited handling of generic types for iOS has plagued us quite a number of times already.

JesusMcCloud:
> That's still removing functionality. You can no longer use `.deserialize<T>()`.
> 
> There should still be a `inline fun <reified T> deserialize`, even if it just forwards to this new method (which is the same thing kotlinx.serialization does).

it is not removing functionality. the current release could only do hardcoded bytearrays. so it is a regression from the previous dev state, but not from the release state. from the release state it, is added functionality

JesusMcCloud:
> > That's still removing functionality. You can no longer use `.deserialize<T>()`.
> > There should still be a `inline fun <reified T> deserialize`, even if it just forwards to this new method (which is the same thing kotlinx.serialization does).
> 
> it is not removing functionality. the current release could only do hardcoded bytearrays. so it is a regression from the previous dev state, but not from the release state. from the release state it, is added functionality

(which is, why no changelog is necessary)

iaik-jheher:
alright, but is there any reason not to offer the reified inline helper?

PR #213, opened 2024-12-03
Feature/2024/12/03 fix bitset iterator

acrusage-iaik:
Fixes the bitset iterator since it was off by 1 (but please double check, the iterator semantics here are new to me)


iaik-jheher (approved):
no, that's correct (and was incorrect), good catch

PR #219, opened 2024-12-10
Fix COSE JSON serialization

nodh:
This might happen when parsing credential responses from OID4VCI in vc-k, see https://github.com/a-sit-plus/vck/commit/cd918f06ff2060fc1c46a265c461843993987a65

Please also release as 3.11.1


n0900 (approved):
small nit

PR #223, opened 2024-12-16
COSE: Store raw payload for signature verification

nodh:
Fixes COSE signature verification as the comment in https://github.com/a-sit-plus/vck/issues/72#issuecomment-2539774230 uncovered.

See also https://github.com/a-sit-plus/vck/pull/180

### DIFF
--- a/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedSerializer.kt
+++ b/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedSerializer.kt
@@ -9,11 +9,7 @@ import kotlinx.serialization.KSerializer
 import kotlinx.serialization.builtins.ByteArraySerializer
 import kotlinx.serialization.cbor.CborEncoder
 import kotlinx.serialization.cbor.ValueTags
-import kotlinx.serialization.descriptors.SerialDescriptor
-import kotlinx.serialization.descriptors.SerialKind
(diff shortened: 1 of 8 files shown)


indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedSerializer.kt
```diff
@@ -1,152 +1,97 @@
...
-                        2,
-                        ByteStringWrapperSerializer(parameterSerializer),
-                        ByteStringWrapper(value.payload!!)
-                    )
-                }
-                encodeSerializableElement(descriptor, 3, ByteArraySerializer(), value.rawSignature)
-            }
+    private fun ByteArray.toTypedPayload(): P =
+        if (parameterSerializer == ByteArraySerializer()) {
+            typed()
```
JesusMcCloud, line 78: please inline this here, since here is the only legal use of such a cast

JesusMcCloud (changes requested):
If you want a point release for this, you'll need to reintroduce the removed c'tor

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedSerializer.kt
nodh: Fixed that

JesusMcCloud (approved):

PR #227, opened 2025-01-10
COSE: Detached Payload

nodh:
Is in line with https://datatracker.ietf.org/doc/rfc9052/ and is required to implement ISO/IEC 18013-7 Annex B (mDoc resonses for OpenID4VP).

### DIFF
--- a/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
+++ b/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
@@ -42,7 +42,9 @@ data class CoseSigned<P : Any?> internal constructor(
     fun prepareCoseSignatureInput(
         externalAad: ByteArray = byteArrayOf(),
         detachedPayload: ByteArray? = null,
-    ): ByteArray = wireFormat.toCoseSignatureInput(externalAad, detachedPayload)
+    ): ByteArray = detachedPayload
+        ?.let { wireFormat.toCoseSignatureInputWithDetachedPayload(externalAad, it) }
+        ?: wireFormat.toCoseSignatureInput(externalAad)
 
     fun serialize(parameterSerializer: KSerializer<P>): ByteArray = coseCompliantSerializer
         .encodeToByteArray(CoseSignedSerializer(parameterSerializer), this)
@@ -139,6 +141,7 @@ data class CoseSigned<P : Any?> internal constructor(
                 ByteStringWrapperSerializer(payloadSerializer),
                 ByteStringWrapper(this)
             ).wrapInCborTag(24)
+
             else -> null
         }
 
--- a/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
+++ b/indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
@@ -39,12 +39,13 @@ data class CoseSigned<P : Any?> internal constructor(
     val wireFormat: CoseSignedBytes,
 ) {
 
+    /**
+     * @param detachedPayload only to be set when [payload] is null, i.e. it is transported externally
+     */
     fun prepareCoseSignatureInput(
         externalAad: ByteArray = byteArrayOf(),
         detachedPayload: ByteArray? = null,
-    ): ByteArray = detachedPayload
-        ?.let { wireFormat.toCoseSignatureInputWithDetachedPayload(externalAad, it) }
-        ?: wireFormat.toCoseSignatureInput(externalAad)
+    ): ByteArray = wireFormat.toCoseSignatureInput(externalAad, detachedPayload)
 
     fun serialize(parameterSerializer: KSerializer<P>): ByteArray = coseCompliantSerializer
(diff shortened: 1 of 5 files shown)


indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
```diff
@@ -39,8 +39,10 @@ data class CoseSigned<P : Any?> internal constructor(
     val wireFormat: CoseSignedBytes,
 ) {
 
-    fun prepareCoseSignatureInput(externalAad: ByteArray = byteArrayOf()): ByteArray =
-        wireFormat.toCoseSignatureInput(externalAad)
+    fun prepareCoseSignatureInput(
+        externalAad: ByteArray = byteArrayOf(),
```
JesusMcCloud, line 43: why is this an empty bytearray and not null?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedBytes.kt
```diff
@@ -32,11 +32,12 @@ data class CoseSignedBytes(
 ) {
     fun toCoseSignatureInput(
         externalAad: ByteArray = byteArrayOf(),
```
JesusMcCloud, line 34: i know it is not new, but still: why is this not null?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSigned.kt
nodh: see below

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedBytes.kt
nodh: Quoting https://www.rfc-editor.org/rfc/rfc9052.html#section-4.4

> The externally supplied data from the application, encoded in a bstr type. If this field is not supplied, it defaults to a zero-length byte string.

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedBytes.kt
```diff
@@ -32,11 +32,12 @@ data class CoseSignedBytes(
...
+        detachedPayload: ByteArray? = null,
     ): ByteArray = CoseSignatureInput(
         contextString = "Signature1",
         protectedHeader = protectedHeader.toZeroLengthByteString(),
         externalAad = externalAad,
-        payload = payload,
+        payload = payload ?: detachedPayload,
```
JesusMcCloud, line 40: This seems off: we have a payload member, but can also use a detached payload, which means we can create a CoseSigned with a payload and then pass a detached payload in this function, which will then be ignored. That is a footgun.

JesusMcCloud (changes requested):

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedBytes.kt
nodh: Changed that

JesusMcCloud: How about this: Instead of passing a detached payload, just make the last parameter a flag:
```kotlin
fun toCoseSignatureInputWithDetachedPayload(
        externalAad: ByteArray = byteArrayOf(),
        detachPayload: Boolean = false,
    ): ByteArray = CoseSignatureInput(
        contextString = "Signature1",
        protectedHeader = protectedHeader.toZeroLengthByteString(),
        externalAad = externalAad,
        payload = if(detachPayload) null else payload,
        ).serialize()
```

No more ambiguities this way.

JesusMcCloud (changes requested):

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSignedBytes.kt
JesusMcCloud: My comment was, of course misleading. The new function should go and the old one should have the parameter:

```kotlin
fun toCoseSignatureInput(
    externalAad: ByteArray = byteArrayOf(),
    detachPayload: Boolean = false,
): ByteArray = CoseSignatureInput(
        contextString = "Signature1",
        protectedHeader = protectedHeader.toZeroLengthByteString(),
        externalAad = externalAad,
        payload = if(detachPayload) null else payload,
        ).serialize()
```

nodh: Refactored as discussed offline

JesusMcCloud (approved):
Pending everything's still green.
Thanks for the out-of-band clarifications!

PR #230, opened 2025-01-16
Release/3.12.1

JesusMcCloud:
* Add COSE object creation with detached payload, i.e. setting a `null` payload in `CoseSigned`, and clients are responsible to transport the payload separately


nodh (approved):
:heart:

PR #247, opened 2025-02-11
X.509 v1

nodh:


JesusMcCloud (approved):
I fixed the bit string issue: ASN.1 Bit String was always correct, but the conversion from/to BitSet was causing trouble (as intended by the test case). So now we keep 'em bit strings raw

PR #255, opened 2025-02-25
fancy social previews

JesusMcCloud:
how to cherry-pick this onto main and keep a linear history given the github buttons don't do that?


indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
```diff
@@ -159,7 +159,8 @@ class ObjectIdentifier @Throws(Asn1Exception::class) private constructor(
         }
 
         /**
-         * Casts out the evil demons that haunt OID components encoded into [rawValue]
+         * Casts out the evil demons that haunt OID components encoded into ASN.1 content [rawValue].
+         * If you want to parse human-readable OID representations, just use the ObjectIdentifier constructor!
```
iaik-jheher, line 163: rename to `parseASN1` instead of having smart kdoc

JesusMcCloud:
@iaik-jheher how to cherry-pick this onto main and keep a linear history given the github buttons don't do that?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/ObjectIdentifier.kt
JesusMcCloud: now aligned with how corresponding methods in integer, etc. are named

JesusMcCloud:
> > @iaik-jheher how to cherry-pick this onto main and keep a linear history given the github buttons don't do that?
> 
> we could only cherry-pick the first commit, because the second changes the API

NVM. the second one fixes something. we'll just not do it

iaik-jheher (changes requested):
image

iaik-jheher (changes requested):
image

iaik-jheher (approved):

PR #257, opened 2025-02-27
add missing ASN.1 Primitives

JesusMcCloud:
* ASN.1 ENUMERATED
* ASN.1 REAL

PR #258, opened 2025-02-27
Feature/real

### LINKED ISSUE #182: ASN.1 Real
If we want to provide a truly universal kotlinx serialization format, we need to cover ASN.1 REAL number encoding and decoding, in all variants supported by the DER. As a first step, just supporting the encoding that maps easiest to Kotlin's `Double`is sufficient.

The real issue here is testing. BC does not support REAL numbers and the online tools also don't implement a parser we could hijack. Even openssl just print `prim: REAL`

JesusMcCloud:
closes #182.

TODO: content encoding separate from tag


JesusMcCloud:
integrated into #257

PR #260, opened 2025-03-05
Wallet Attestation

nodh:
See https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-key-attestations and https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-wallet-attestations-in-jwt-


JesusMcCloud (commented):
why don't i see a `cnf` property?

nodh:
> why don't i see a `cnf` property?

In the `JsonWebToken`, there is a `cnf` property, unchanged

JesusMcCloud (approved):

JesusMcCloud:
my bad

PR #272, opened 2025-04-09
remove debug-only serialization from cryptographic datatypes

JesusMcCloud:
symmetric encryption and rsa or hmac separation already introduces breaking changes all over the place. While we're at it, we can go ahead and break even more stuff. better now than a separate breaking release

### DIFF
--- a/indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Integer.kt
+++ b/indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Integer.kt
@@ -35,7 +35,7 @@ fun Asn1Integer(number: ULong) =
  * It has a [sign] though, and supports [twosComplement] representation and converting [fromTwosComplement].
  * Hence, it directly interoperates with [Kotlin MP BigNum](https://github.com/ionspin/kotlin-multiplatform-bignum) and the JVM BigInteger.
  */
-@Serializable(with = Asn1IntegerSerializer::class)
+@Serializable(with = Asn1IntegerSerializer::class) //this here stays, as it might be useful in general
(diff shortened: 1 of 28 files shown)


indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Integer.kt
```diff
@@ -35,7 +35,7 @@ fun Asn1Integer(number: ULong) =
  * It has a [sign] though, and supports [twosComplement] representation and converting [fromTwosComplement].
  * Hence, it directly interoperates with [Kotlin MP BigNum](https://github.com/ionspin/kotlin-multiplatform-bignum) and the JVM BigInteger.
  */
-@Serializable(with = Asn1IntegerSerializer::class)
+@Serializable(with = Asn1IntegerSerializer::class) //this here stays, as it might be useful in general
```
nodh, line 38: Why just for `Asn1Integer`s?

nodh (approved):

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Integer.kt
JesusMcCloud: good point! should also cover the real type!

JesusMcCloud: done

JesusMcCloud: Because the numeric types might be generally usable if you want arbitrary-precision numbers

iaik-jheher (approved):

PR #279, opened 2025-04-16
Improve PEM exception text on incorrect usage

iaik-jheher:
So if you try to fit the proverbial square peg in the proverbial round hole, we non-proverbially tell you what you're doing wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 3.0
 
+### NEXT
+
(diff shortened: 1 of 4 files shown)


JesusMcCloud:
but this will catch the PEM boundaries for types that have been instantiated before a mismatch occurs, right?!

iaik-jheher:
cf. whyoleg/sweet-spi#10

Issue #180, opened 2024-10-30
Provide COSE/JOSE-Supreme helper modules


JesusMcCloud:
we want extensions to directly support what is described in the examples OOTB

PR #281, opened 2025-04-24
Improve examples

JesusMcCloud:


docs/docs/examples.md
```diff
@@ -43,6 +43,36 @@ The reason for keeping this fourth parameter is convenience and efficiency: For
...
+val jwsAlgorithm = jwsObject.header.algorithm
+//JWS is very permissive, so we need to check that the alg makes sense
+require(jwsAlgorithm is JwsAlgorithm.Signature) { "Algorithm not supported: $jwsAlgorithm" }
+val key = jwsObject.header.publicKey ?: pubKeyFromTrustedSource
+
+//TODO: check whether key is trusted by whatever means apply in your case
+
+val verifier = jwsAlgorithm.verifierFor(publicKey).getOrThrow()
+//Verify cryptographically
+val verified = verifier.verify(jwsObject.plainSignatureInput, jwsObject.signature).isSuccess
```
iaik-jheher, line 71: why is there no extension function for this

docs/docs/examples.md
```diff
@@ -43,6 +43,36 @@ The reason for keeping this fourth parameter is convenience and efficiency: For
...
+val jwsObject = JwsSigned.deserialize(jws)
+//check if a sensible algorithm is set
+val jwsAlgorithm = jwsObject.header.algorithm
+//JWS is very permissive, so we need to check that the alg makes sense
+require(jwsAlgorithm is JwsAlgorithm.Signature) { "Algorithm not supported: $jwsAlgorithm" }
+val key = jwsObject.header.publicKey ?: pubKeyFromTrustedSource
+
+//TODO: check whether key is trusted by whatever means apply in your case
+
+val verifier = jwsAlgorithm.verifierFor(publicKey).getOrThrow()
```
iaik-jheher, line 68: what is `publicKey`, and how is it different from `key`?

docs/docs/examples.md
```diff
@@ -43,6 +43,36 @@ The reason for keeping this fourth parameter is convenience and efficiency: For
...
+5. Check timing-related and other constraints
+
+```kotlin
+//parse serialized JWS
+val jwsObject = JwsSigned.deserialize(jws)
+//check if a sensible algorithm is set
+val jwsAlgorithm = jwsObject.header.algorithm
+//JWS is very permissive, so we need to check that the alg makes sense
+require(jwsAlgorithm is JwsAlgorithm.Signature) { "Algorithm not supported: $jwsAlgorithm" }
+val key = jwsObject.header.publicKey ?: pubKeyFromTrustedSource
```
iaik-jheher: should this not check for the same type as algorithm type as `key`?

ED: After validating `key`? The order of operations is extremely relevant here.

docs/docs/examples.md
JesusMcCloud: see #180

docs/docs/examples.md
JesusMcCloud: should not the verifier creation simply fail if it does not check out?

docs/docs/examples.md
JesusMcCloud: fixed, thanks!

docs/docs/examples.md
JesusMcCloud: edited to explicitly enforce a source of trusted keys

iaik-jheher (approved):

PR #282, opened 2025-04-29
correct type bounds for base signing provider

iaik-jheher:
`PlatformSigningProvider` should require `PlatformSigningKeyConfigurationBase`, but `SigningProvider` should only require `SigningProviderSigningKeyConfigurationBase`.


JesusMcCloud (approved):
seems legit

JesusMcCloud (changes requested):
Well, if the tests can be made to work again, I'll re-review

PR #289, opened 2025-05-06
update to latest conventions with unmanaged Kotlin version

JesusMcCloud:
updates every dependency and uses the new conventions plugin with a freely-definable Kotlin version

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,12 +2,15 @@ import org.jetbrains.dokka.gradle.DokkaMultiModuleTask
 
 plugins {
     id("at.asitplus.gradle.conventions") version "20250628"
-    id("io.kotest.multiplatform") version (System.getenv("KOTEST_VERSION_ENV")?.let { it.ifBlank { null } }
-        ?: libs.versions.kotest.get())
-    kotlin("multiplatform") version (System.getenv("KOTLIN_VERSION_ENV")?.let { it.ifBlank { null } }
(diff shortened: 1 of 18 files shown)


build.gradle.kts
```diff
@@ -1,7 +1,13 @@
 import org.jetbrains.dokka.gradle.DokkaMultiModuleTask
 
 plugins {
-    id("at.asitplus.gradle.conventions") version "2.1.20+20250324"
+    id("at.asitplus.gradle.conventions") version "20250628"
+    id("io.kotest.multiplatform") version (System.getenv("KOTEST_VERSION_ENV")?.let { it.ifBlank { null } }
+        ?: libs.versions.kotest.get())
+    kotlin("multiplatform") version (System.getenv("KOTLIN_VERSION_ENV")?.let { it.ifBlank { null } }
```
iaik-jheher: isn't there `takeIf` or something like that

iaik-jheher: also can we define a `val kotlinVersion` somewhere further up instead of copy/pasting this snippet

JesusMcCloud: nah, the version catalog won't be ready by then. but i should mos def simplify the expressions

JesusMcCloud:
> i would still prefer if we found a way not to copy/paste the same expression twice, but here you go

HA! if i declare it inside the `plugins` block, it works!

iaik-jheher (approved):

PR #297, opened 2025-06-03
Feature/certificate validation

JesusMcCloud:
Created now for me to be able to comment


indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -282,7 +311,34 @@ data class X509Certificate @Throws(IllegalArgumentException::class) constructor(
 
     val publicKey: CryptoPublicKey get() = tbsCertificate.publicKey
 
-    companion object : PemDecodable<Asn1Sequence, X509Certificate>(EB_STRINGS.DEFAULT, EB_STRINGS.LEGACY) {
+    fun hasReplayingExtensions(): Boolean =
```
JesusMcCloud: this should never be happening in the first place

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -9,7 +9,7 @@ import kotlinx.serialization.Serializable
  * X.509 Certificate Extension
  */
 @ConsistentCopyVisibility
-data class X509CertificateExtension @Throws(Asn1Exception::class) private constructor(
+data class X509CertificateExtension @Throws(Asn1Exception::class) constructor(
```
JesusMcCloud: how did we instantiate before?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/BasicConstraints.kt
```diff
@@ -0,0 +1,44 @@
...
+import at.asitplus.signum.indispensable.asn1.KnownOIDs
+import at.asitplus.signum.indispensable.asn1.encoding.Asn1
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToBoolean
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToInt
+import at.asitplus.signum.indispensable.asn1.encoding.encodeToAsn1Primitive
+import at.asitplus.signum.indispensable.pki.X509CertificateExtension
+
+class BasicConstraints(
+    val ca: Boolean = false,
+    val pathLenConstraint: Int
```
JesusMcCloud: not max value by default, but a nullable Int

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/BasicConstraints.kt
```diff
@@ -0,0 +1,44 @@
...
+        +ca.encodeToAsn1Primitive()
+        +pathLenConstraint.encodeToAsn1Primitive()
+    }
+
+    companion object : Asn1Decodable<Asn1Sequence, BasicConstraints> {
+        override fun doDecode(src: Asn1Sequence): BasicConstraints {
+            var ca = false
+            var pathLenConstraint: Int = Int.MAX_VALUE
+            if (src.children.isNotEmpty()) ca = src.children[0].asPrimitive().decodeToBoolean()
+            if (src.children.size > 1) pathLenConstraint =
```
JesusMcCloud: what if more children exist?

iaik-jheher:
this PR diff looks tainted; there is stuff like the 3.16.1 -> 3.17.0 snapshot change in there? @JesusMcCloud please double check that the diff is valid before i review

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/BERTags.kt
```diff
@@ -46,6 +46,9 @@ object BERTags {
     const val DURATION: UByte = 0x22u
     const val OBJECT_IDENTIFIER_IRI: UByte = 0x23u
     const val RELATIVE_OID_IRI: UByte = 0x24u
+    const val RFC822_NAME: UByte = 1u
+    const val DNS_NAME: UByte = 2u
+    const val URI_NAME: UByte = 6u
```
iaik-jheher, line 51: it feels like this pull request has a lot of miscellaneous changes that are not directly related to the stated PR purpose; can we split these out into a separate PR to keep the review change set manageable?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/BitSet.kt
```diff
@@ -116,7 +116,7 @@ class BitSet private constructor(private val buffer: MutableList<Byte>) : Iterab
      * This is the real deal, as it has [Long] indices
      */
     inline fun forEachIndexed(block: (i: Long, it: Boolean) -> Unit) {
-        for (i in 0..<length()) block(i, this[i])
+        for (i in 0..<length()) block(i, this.get(i))
```
iaik-jheher: remove this fluff change please

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/BitSet.kt
```diff
@@ -205,7 +205,7 @@ class BitSet private constructor(private val buffer: MutableList<Byte>) : Iterab
         if (other::class != BitSet::class) return false
         other as BitSet
         forEachIndexed { i, it ->
-            if (other[i] != it) return false
+            if (other.get(i) != it) return false
```
iaik-jheher: same

indispensable-asn1/src/jvmTest/kotlin/at/asitplus/signum/indispensable/BitSetTest.kt
```diff
@@ -161,7 +161,7 @@ class BitSetTest : FreeSpec({
 
                 input.forEachIndexed { i, b ->
                     withClue("jvm[$i]") { jvm[i] shouldBe b }
-                    withClue("kmm[$i]") { kmm[i.toLong()] shouldBe b }
+                    withClue("kmm[$i]") { kmm.get(i.toLong()) shouldBe b }
```
iaik-jheher: same, why is this here

indispensable-asn1/src/jvmTest/kotlin/at/asitplus/signum/indispensable/BitSetTest.kt
```diff
@@ -235,7 +235,7 @@ class BitSetTest : FreeSpec({
 
                     input.forEachIndexed { i, b ->
                         withClue("jvm[$i]") { jvm[i] shouldBe b }
-                        withClue("kmm[$i]") { kmm[i.toLong()] shouldBe b }
+                        withClue("kmm[$i]") { kmm.get(i.toLong()) shouldBe b }
```
iaik-jheher: same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -76,6 +87,16 @@ sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
...
+    class EmailAddress(override val value: Asn1Element) : AttributeTypeAndValue() {
+        override val oid = OID
+
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+
+        companion object {
+            val OID = KnownOIDs.emailAddress_1_2_840_113549_1_9_1
+        }
+    }
+
```
iaik-jheher, line 104: this also seems unrelated to certificate chain validation

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -94,7 +115,16 @@ sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
...
+        val thisStr = (this.value as? Asn1Primitive)?.asAsn1String()?.value
+        val otherStr = (other.value as? Asn1Primitive)?.asAsn1String()?.value
+        val thisNormalized = thisStr?.replace("\\s+".toRegex(), "")?.lowercase()
+        val otherNormalized = otherStr?.replace("\\s+".toRegex(), "")?.lowercase()
+        if (thisStr != null && otherStr != null) {
+            if (thisNormalized != otherNormalized) return false
+        } else {
+            if (value != other.value) return false
+        }
+
```
iaik-jheher, line 132: i am uncertain what this change is doing; how is it related to certificate validation?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -30,18 +56,21 @@ constructor(
...
     val issuerUniqueID: Asn1BitString? = null,
     val subjectUniqueID: Asn1BitString? = null,
     val extensions: List<X509CertificateExtension>? = null,
 ) : Asn1Encodable<Asn1Sequence> {
 
     init {
-        if (extensions?.distinctBy { it.oid }?.size != extensions?.size) throw Asn1StructuralException("Multiple extensions with the same OID found")
+        if (extensions?.distinctBy { it.oid }?.size != extensions?.size) throw Asn1StructuralException(
+            "Multiple extensions with the same OID found"
+        )
```
iaik-jheher:
```suggestion
        if (extensions?.distinctBy { it.oid }?.size != extensions?.size)
            throw Asn1StructuralException("Multiple extensions with the same OID found")
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -63,7 +92,20 @@ constructor(
...
+        get() = extensions
+            ?.find { it.oid == KnownOIDs.keyUsage }
+            ?.value
+            ?.asEncapsulatingOctetString()
+            ?.children
+            ?.getOrNull(0)
+            ?.let { it as? Asn1Primitive }
+            ?.let(Asn1BitString::decodeFromTlv)
+            ?.let(X509KeyUsage::doDecode)
+            ?: emptySet()
```
iaik-jheher: are we sure this isn't too permissive, by just falling back to the empty set (instead of throwing) on any unexpected behavior? For example, is it valid for the thing to be present but not be an Asn1Primitive?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -72,14 +114,14 @@ constructor(
             version?.let { +Version(it) }
             +Asn1Primitive(Asn1Element.Tag.INT, serialNumber)
             +signatureAlgorithm
-            +Asn1.Sequence { issuerName.forEach { +it } }
+            +issuerName
```
iaik-jheher, line 126: these seem functionally different; is there a bug in our current implementation? if yes, why did it not surface earlier?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -72,14 +114,14 @@ constructor(
...
 
             +Asn1.Sequence {
                 +validFrom
                 +validUntil
             }
 
-            +Asn1.Sequence { subjectName.forEach { +it } }
+            +subjectName
```
iaik-jheher, line 133: these also seem functionally different

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -282,7 +329,32 @@ data class X509Certificate @Throws(IllegalArgumentException::class) constructor(
 
     val publicKey: CryptoPublicKey get() = tbsCertificate.publicKey
 
-    companion object : PemDecodable<Asn1Sequence, X509Certificate>(EB_STRINGS.DEFAULT, EB_STRINGS.LEGACY) {
+    fun isSelfIssued(): Boolean = tbsCertificate.subjectName == tbsCertificate.issuerName
+
+    inline fun <reified T : X509CertificateExtension> findExtension(): T? {
+        return this.tbsCertificate.extensions?.firstOrNull { it is T } as? T
```
iaik-jheher:
```suggestion
        return this.tbsCertificate.extensions?.firstNotNullOfOrNull { it as? T }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -320,9 +393,19 @@ private
 /** De-/serializes Base64 strings to/from [ByteArray] */
 object Asn1BitStringSerializer : TransformingSerializerTemplate<Asn1BitString?, String>(
     parent = String.serializer(),
-    encodeAs = { if (it == null) "" else byteArrayOf(it.numPaddingBits, *it.rawBytes).encodeToString(Base64Strict) },
+    encodeAs = {
+        if (it == null) "" else byteArrayOf(
+            it.numPaddingBits,
+            *it.rawBytes
+        ).encodeToString(Base64Strict)
+    },
```
iaik-jheher: please don't add unnecessary whitespace changes to a pull request, they just bloat the review diff

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -3,22 +3,22 @@ package at.asitplus.signum.indispensable.pki
...
+import at.asitplus.signum.indispensable.pki.pkiExtensions.NameConstraintsExtension
+import at.asitplus.signum.indispensable.pki.pkiExtensions.PolicyConstraintsExtension
+import at.asitplus.signum.indispensable.pki.pkiExtensions.PolicyMappingsExtension
 
 /**
  * X.509 Certificate Extension
  */
-@ConsistentCopyVisibility
-data class X509CertificateExtension @Throws(Asn1Exception::class) private constructor(
+open class X509CertificateExtension @Throws(Asn1Exception::class) private constructor(
```
iaik-jheher, line 32: this being `open` instead of `sealed` makes me immediately suspicious

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -3,22 +3,22 @@ package at.asitplus.signum.indispensable.pki
...
     override val oid: ObjectIdentifier,
-    val value: Asn1Element,
-    val critical: Boolean = false
+    open val value: Asn1Element,
+    open val critical: Boolean = false
```
iaik-jheher: why are these `open val`s instead of the constructor being `protected` if you expect it to be invoked from subclasses

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -3,22 +3,22 @@ package at.asitplus.signum.indispensable.pki
...
 ) : Asn1Encodable<Asn1Sequence>, Identifiable {
 
-    init {
-        if (value.tag != Asn1Element.Tag.OCTET_STRING) throw Asn1TagMismatchException(Asn1Element.Tag.OCTET_STRING, value.tag)
-    }
-
```
iaik-jheher: is this check not correct?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -42,11 +42,28 @@ data class X509CertificateExtension @Throws(Asn1Exception::class) private constr
...
+                else -> decodeBase(src)
+            }
+        }
+
+        @Throws(Asn1Exception::class)
+        fun decodeBase(src: Asn1Sequence): X509CertificateExtension {
+            val id = src.nextChild().asPrimitive().readOid()
             val critical =
-                if (src.children[1].tag == Asn1Element.Tag.BOOL) (src.children[1] as Asn1Primitive).content[0] == 0xff.toByte() else false
+                if (src.children[1].tag == Asn1Element.Tag.BOOL) src.nextChild().asPrimitive().content[0] == 0xff.toByte() else false
```
iaik-jheher: how do we handle unknown critical extensions? (please confirm that the validator checks this)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509KeyUsage.kt
```diff
@@ -0,0 +1,32 @@
...
+    KEY_AGREEMENT(4),
+    KEY_CERT_SIGN(5),
+    CRL_SIGN(6),
+    ENCIPHER_ONLY(7),
+    DECIPHER_ONLY(8)
+
+    ;
+
+    companion object {
+        fun doDecode(encodedValue: Asn1BitString): Set<X509KeyUsage> {
```
iaik-jheher: `parseExtension` or something?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/BasicConstraintsExtension.kt
```diff
@@ -0,0 +1,57 @@
...
+import at.asitplus.signum.indispensable.pki.X509CertificateExtension
+
+/**
+ * Basic Constraints Extension
+ * RFC 5280: 4.2.1.9.
+*/
+class BasicConstraintsExtension(
+    override val oid: ObjectIdentifier,
+    override val critical: Boolean,
+    override val value: Asn1EncapsulatingOctetString,
```
iaik-jheher: why are these override vals if they are passed into the base class ctor anyway?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
iaik-jheher: (i'm not saying it's not a valid change; but it would be easier if we split those "misc changes" into a separate independent PR first, then did the certificate validation PR on top once that is merged)

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,172 @@
...
+    KnownOIDs.cRLDistributionPoints_2_5_29_31,
+    KnownOIDs.issuingDistributionPoint_2_5_29_28,
+    KnownOIDs.deltaCRLIndicator,
+    KnownOIDs.policyConstraints_2_5_29_36,
+    KnownOIDs.basicConstraints_2_5_29_19,
+    KnownOIDs.subjectAltName_2_5_29_17,
+    KnownOIDs.nameConstraints_2_5_29_30
+)
+
+sealed interface Validator {
```
wkornewald: Could this be made non-sealed, so we can define our own validators outside of this lib?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,172 @@
...
+    fun check(currCert: X509Certificate)
```
wkornewald: Could this be made suspend, so it's more flexible and can e.g. do requests if necessary?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/BasicConstraintsValidator.kt
```diff
@@ -0,0 +1,42 @@
+package at.asitplus.signum.supreme.validate
+
+import at.asitplus.signum.BasicConstraintsException
+import at.asitplus.signum.indispensable.pki.X509Certificate
+import at.asitplus.signum.indispensable.pki.pkiExtensions.BasicConstraintsExtension
+
+class BasicConstraintsValidator(
```
wkornewald: Could the validators which don't depend on any crypto APIs be moved to the indispensable module, so one can combine the certificate code with a different crypto API (e.g. cryptography-kotlin)?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
wkornewald: Also, it would be nice to call this X509Validator or CertificateValidator to distinguish it from other kinds of validators.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
wkornewald: Since critical extensions must all be checked it would be better to return or somehow collect the set of checked extensions for each check() call, so some outer validator (which gets a list of inner validators) can ensure that all critical extensions have actually been checked. Bouncy Castle does something similar, for example.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
StjepanovicSrdjan: Previously, issuerName was a `List<RelativeDistinguishedName>`. I have now refactored it to use the X500Name class, which internally contains a `List<RelativeDistinguishedName>` and performs the same encoding as before.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
StjepanovicSrdjan: Same as [here](https://github.com/a-sit-plus/signum/pull/297#discussion_r2161145579)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: This change was made because, during name chaining validation, there are cases where the chain is considered valid despite differences in whitespace or capitalization.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
StjepanovicSrdjan: Validator checks that, but it should be refactored as stated in the [comment](https://github.com/a-sit-plus/signum/pull/297#discussion_r2160010329) by @wkornewald

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
StjepanovicSrdjan: Good point, I think the presence of the extension is not mandatory, but I should definitely add some checks for stricter parsing when it is present.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -282,7 +312,38 @@ data class X509Certificate @Throws(IllegalArgumentException::class) constructor(
 
     val publicKey: CryptoPublicKey get() = tbsCertificate.publicKey
 
-    companion object : PemDecodable<Asn1Sequence, X509Certificate>(EB_STRINGS.DEFAULT, EB_STRINGS.LEGACY) {
+    val criticalExtensionOids: MutableSet<ObjectIdentifier>
+        get() = this.tbsCertificate.extensions
+            ?.filter { it.critical }
+            ?.mapTo(mutableSetOf()) { it.oid }
+            ?: mutableSetOf()
```
wkornewald: X509Certificate instances should be immutable otherwise you can end up with inconsistent state. For validation you can use `toMutableSet()` and safely work on your internal copy instead.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,156 @@
...
+import at.asitplus.signum.supreme.sign.verifierFor
+import at.asitplus.signum.supreme.sign.verify
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
+
+val uncheckedCriticalExtensionOids = setOf(
+    KnownOIDs.cRLDistributionPoints_2_5_29_31,
+    KnownOIDs.issuingDistributionPoint_2_5_29_28,
+    KnownOIDs.deltaCRLIndicator,
+)
```
wkornewald: Why should we allow certian critical extensions to be unchecked/ignored? Isn't that against the spec?

StjepanovicSrdjan: Yes, you're right that was completely against the spec. It was only a temporary workaround. However, this has now been properly fixed in the latest commit.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -39,19 +61,47 @@ data class X509CertificateExtension @Throws(Asn1Exception::class) private constr
...
+            KnownOIDs.policyConstraints_2_5_29_36 to PolicyConstraintsExtension::decodeFromTlv,
+            KnownOIDs.certificatePolicies_2_5_29_32 to CertificatePoliciesExtension::decodeFromTlv,
+            KnownOIDs.policyMappings to PolicyMappingsExtension::decodeFromTlv,
+            KnownOIDs.inhibitAnyPolicy to InhibitAnyPolicyExtension::decodeFromTlv,
+            KnownOIDs.keyUsage to KeyUsageExtension::decodeFromTlv
+        )
+        private val mutex = Mutex()
+
+
+        suspend fun registerExtensionDecoder(
```
wkornewald: Instead of making this suspend and using a mutex you could use a `MutableStateFlow` with `.update { ... }` to atomically update an immutable map.

JesusMcCloud: This sounds like a pattern we could use in multiple locations, given we'll need this kind of extensibility in other places as well.

PR #305, opened 2025-06-18
Release 3.16.3 / 0.8.3 indispensable-only Hotfix

JesusMcCloud:
* More targets for indispensable modules:
    * watchosSimulatorArm64
    * watchosX64
    * watchosArm32
    * watchosArm64
    * androidNativeX64
    * androidNativeX86
    * androidNativeArm32
    * androidNativeArm64
* Drop OKIO dependency from `indispensable-josef` which was only ever used for to compute a SHA-256 thumbprint and replace it by a pure Kotlin SHA-256 implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## 3.0
 
+### 3.16.3 / 0.8.3 indispensable-only Hotfix
+* More targets:
+    * watchosSimulatorArm64
+    * watchosX64
+    * watchosArm32
+    * watchosArm64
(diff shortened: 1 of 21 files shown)


BernatCarbo:
Can `watchosDeviceArm64` not be included?

JesusMcCloud:
> consider adding the extra target for asn.1 though since it has no dependencies (not a blocker imo, but i don't see a downside)

I can do that, its just a bit fiddly, because the test sources won't compile then. should be fine for a release build, though

PR #306, opened 2025-06-23
Fix Jose cert serialization

JesusMcCloud:
* Fix erroneous Base64URL encoding in JOSE data classes, adding missing serializers in addition to Base64Url encoding:
    * `X509CertificateBase64Serializer`
    * `CertificateChainBase64Serializer`

@tlenz: No other places except for attestation use the base64 url serializer (according to intelliJ). please double-check!
I'll then directly push this into the 3.16.3 release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## 3.0
 
 ### 3.16.3 / 0.8.3 indispensable-only Hotfix
+* Fix erroneous Base64URL encoding in JOSE data classes
+    * `toString()` of `X509Certificate` and `TbsCertificate` have also been adapted to use Base64 Strict
+* Add missing serializers in addition to Base64Url encoding:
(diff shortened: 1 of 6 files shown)


tlenz (approved):
JWS / JWE and JsonWebKey data-classes looks good

tlenz:
However, I inspect the original commit https://github.com/a-sit-plus/signum/commit/3a6690e398efe60f32908fb9f2a91db733c849f5 and there, the line: https://github.com/a-sit-plus/signum/blob/3a6690e398efe60f32908fb9f2a91db733c849f5/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt#L139C34-L139C51
looks also problematic because it looks like PEM encoding but does not use Base64

JesusMcCloud:
For the sake of completeness: I also botched the VC-K tests. VC-K release will follow

PR #312, opened 2025-07-02
Enhancement/Decoding X509Certificates with unsupported X509SignatureAlgorithm

StjepanovicSrdjan:
Besides storing rawSignature in X509Certificate, a similar approach is used in tbsCertificate for storing rawPublicKey. This helps avoid decoding failures in certificates that use, for example, DSA keys.


JesusMcCloud:
please amend the changelog. I'll review later

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -500,6 +500,11 @@ sealed class Asn1Structure(
      */
     fun peek() = if (!hasMoreChildren()) null else children[index]
 
+    /**
+     * Consumes all remaining children
+     * */
+    fun consumeRemainingChildren() { index = children.size }
```
JesusMcCloud: but that just advances the index and discards everythin?!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -15,28 +27,22 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 
 @Serializable(with = X509SignatureAlgorithmSerializer::class)
-enum class X509SignatureAlgorithm(
+sealed class X509SignatureAlgorithm(
```
JesusMcCloud, line 22: I am very much against a sealed class and "other" [just as Raymond Chen](https://devblogs.microsoft.com/oldnewthing/20250217-00/?p=110873).

Make it `open`, so the Other stuff just disappears into the base class and introduce the ability to register other algorithms. The fact that this will be failing to convert to `SigantureAlgorithm` in other paces is fine for now.

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: When decoding an unsupported X509SignatureAlgorithm, we don’t need to consume the ASN.1 structure; instead, we just store it as a value. This causes the ASN.1 trailing exception to occur. ConsumeRemainingChildren() serves as a workaround for calling nextChild() repeatedly until the structure is fully parsed. Therefore, this function is called only for algorithms that we don’t know how to parse.

JesusMcCloud: I don't get it. we're not doing any decoding operation on the ASN.1 element in question, so we don't need to touch the element or its child nodes at all, so there shouldn't be any checks about training elements anyhow?

JesusMcCloud: I also don't feel this function should even be a thing. it is just ignoring content and thus misleading. If (at all) it should be `discardRemainignChildren`, but I'd need convincing of a use case for that

StjepanovicSrdjan: If you check the doDecode function in  X509SignatureAlgorithm:
```@Throws(Asn1Exception::class)
        override fun doDecode(src: Asn1Sequence): X509SignatureAlgorithm = runRethrowing {
            when (val oid = (src.nextChild() as Asn1Primitive).readOid()) {
                KnownOIDs.rsaPSS -> parsePssParams(src)
                else -> {
                    val alg = fromOid(oid)
                    if (alg is RSA) {
                        val tag = src.nextChild().tag
                        if (tag != Asn1Element.Tag.NULL)
                            throw Asn1TagMismatchException(Asn1Element.Tag.NULL, tag, "RSA Params not allowed.")
                    }
                    alg ?: Other(oid, src).also { src.consumeRemainingChildren() }
                }
            }
        }
```
        
This is where I use that function. For DSA alg, the same handling as for RSA is required—parsing the next child, which should have the NULL tag to indicate that there are no parameters. If I don't do that, or if I don't call consumeRemainingChildren(), there will be trailing children left, which will cause an error.

The reason for me not to do the same thing as for RSA is that this is the case for specific algorithm and since we don't support it, I assume that we shouldn't implement parsing specifics for that sole alg.

iaik-jheher: is the NULL-tag-to-indicate-last-child something that is specified generically for any algorithm in the RFC?

StjepanovicSrdjan: ``AlgorithmIdentifier  ::=  SEQUENCE  {
        algorithm               OBJECT IDENTIFIER,
        parameters              ANY DEFINED BY algorithm OPTIONAL  }``

"The contents of the optional parameters
   field will vary according to the algorithm identified." [RFC-4.1.1.2](https://www.rfc-editor.org/rfc/rfc5280#section-4.1.1.2)
   
In the RSA signature algorithm, there should always be a NULL tag, whereas for EC algorithms, only the OID should be present without any additional parameters. Also, if I understood the RFC correctly, DSA shouldn't have any additional parameters either. In my opinion, the `consumeRemainingChildren()` function makes sense because the "parameters" field in the ASN.1 structure is OPTIONAL.

iaik-jheher: In this case, I then agree that for unknown algorithms, any parameters are permitted. They should still be consumed and represented as a list/sequence/whatever, though.

StjepanovicSrdjan: Updated X509SignatureAlgorithm to store list of parameters as Asn1Elements, allowing me to remove discussed consumeRemainingChildren()" function and just parse remaining children as parameters

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -204,6 +257,6 @@ object X509SignatureAlgorithmSerializer : KSerializer<X509SignatureAlgorithm> {
 
     override fun deserialize(decoder: Decoder): X509SignatureAlgorithm {
         val decoded = decoder.decodeString()
-        return X509SignatureAlgorithm.entries.first { it.name == decoded }
+        return X509SignatureAlgorithm.registeredAlgorithms.first { it.name == decoded }
```
JesusMcCloud: good thing I have a PR going that gets rid of this ugly hack ;-). fine here, of course

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -171,7 +199,7 @@ constructor(
                 RelativeDistinguishedName.decodeFromTlv(it as Asn1Set)
             }
 
-            val cryptoPublicKey = CryptoPublicKey.decodeFromTlv(src.nextChild() as Asn1Sequence)
+            val rawKey = src.nextChild() as Asn1Sequence
```
JesusMcCloud: there's a shorthand `.asSequence` but this is just cosmetic.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -235,9 +263,10 @@ val CryptoSignature.x509Encoded
  * - EC is DER-encoded then wrapped in a bit string
  */
 fun CryptoSignature.Companion.fromX509Encoded(alg: X509SignatureAlgorithm, it: Asn1Primitive) =
-    when (alg.isEc) {
-        true -> CryptoSignature.EC.decodeFromDer(it.asAsn1BitString().rawBytes)
-        false -> CryptoSignature.RSA.decodeFromTlv(it)
+    when (alg) {
+        is X509SignatureAlgorithm.EC -> CryptoSignature.EC.decodeFromDer(it.asAsn1BitString().rawBytes)
+        is X509SignatureAlgorithm.RSA -> CryptoSignature.RSA.decodeFromTlv(it)
+        else -> throw IllegalArgumentException("Unsupported signature algorithm.")
```
JesusMcCloud, line 275: maybe leave a comment here so that we don't miss updating this when we make our core data classes equally extensible?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -91,35 +94,85 @@ enum class X509SignatureAlgorithm(
...
         }
 
     override val algorithm: SignatureAlgorithm
         get() = when (this) {
-            ES256, ES384, ES512 -> SignatureAlgorithm.ECDSA(this.digest, null)
-            PS256, PS384, PS512 -> SignatureAlgorithm.RSA(this.digest, RSAPadding.PSS)
-            RS1, RS256, RS384, RS512 -> SignatureAlgorithm.RSA(this.digest, RSAPadding.PKCS1)
+            is EC -> SignatureAlgorithm.ECDSA(digest, null)
+            is RSA -> SignatureAlgorithm.RSA(digest, if (pssBits != null) RSAPadding.PSS else RSAPadding.PKCS1)
+            else -> throw IllegalArgumentException("Unsupported signature algorithm.")
```
JesusMcCloud: maybe leave a comment here so that we don't miss updating this when we make our core data classes equally extensible?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -91,35 +94,85 @@ enum class X509SignatureAlgorithm(
...
+            )
+        )
+        val registeredAlgorithms: Set<X509SignatureAlgorithm>
+            get() = _registeredAlgorithms.value
 
-        @Throws(Asn1OidException::class)
-        private fun fromOid(oid: ObjectIdentifier) = catching { entries.first { it.oid == oid } }.getOrElse {
-            throw Asn1OidException("Unsupported OID: $oid", oid)
+        fun register(algorithm: X509SignatureAlgorithm) {
+            _registeredAlgorithms.update { it + algorithm}
```
JesusMcCloud: We will need this exact pattern in many other places in the near future. I think we should seize the opportunity and introduce an interface (with default implementations of methods, where possible) that encapsulates this functionality. @iaik-jheher do you agree?

JesusMcCloud (changes requested):
Looking good, but I think we can push this further and encapsulate the newly introduced pattern into an interface. Let's wait for comments form @iaik-jheher, before actually doing it.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
StjepanovicSrdjan: I usually use that shorthand, but since the rest of the function use `as Asn1Something`  I followed the pattern to keep it consistent.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
StjepanovicSrdjan: Here, I use a set to store registered algorithms, but in the interface, it would make more sense to use a map. For example, in certificate validation, I use a Map with the OID as the key to register new X509Extensions. It would also be easier here to use a map with the OID as the key and access values directly by OID instead of using the fromOid method. However, that wasn’t possible because RSA-PSS algorithms use the same OID regardless of the number of bits.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
StjepanovicSrdjan: added in [75e639e](https://github.com/a-sit-plus/signum/pull/312/commits/75e639eb5481ea49dee0cdf9d1d4e304ce537b76)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
StjepanovicSrdjan: added in [75e639e](https://github.com/a-sit-plus/signum/pull/312/commits/75e639eb5481ea49dee0cdf9d1d4e304ce537b76)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
StjepanovicSrdjan: Introduced Registry interface in [ac90ceb](https://github.com/a-sit-plus/signum/pull/312/commits/ac90ceb419b8225f858f2fc8ce6975dcecb05ea8). Tried to make it as generic as possible by storing a Map of registered items, with the ability for subclasses to define how the key is extracted from each item. Usually that can be OID (for example X509CertificateExtension in #297 ) but in the X509SignatureAlgorithm case that wasn't possible since RSA-PSS use the same OID regardless of the number of bits.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/Registry.kt
```diff
@@ -0,0 +1,37 @@
...
+    fun register(item: T, key: K? = null)
+}
+
+/**
+ * A mutable implementation of [Registry] backed by a [MutableStateFlow].
+ * This abstract class allows runtime extension of the registry and optionally supports automatic
+ * key extraction from the item using the [extractKey] function.
+ */
+abstract class MutableRegistry<K, T>(
+    private val extractKey: ((T) -> K)? = null
```
JesusMcCloud: not limited to a key, but a generic transformation, so I'd rename this.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -15,28 +28,18 @@ import kotlinx.serialization.encoding.Decoder
 import kotlinx.serialization.encoding.Encoder
 
 @Serializable(with = X509SignatureAlgorithmSerializer::class)
-enum class X509SignatureAlgorithm(
+open class X509SignatureAlgorithm private constructor(
```
JesusMcCloud: the constructor should be public to allow for constructing some "unknown" signature algorithms, if needed for passing them along without accessing them

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/Registry.kt
```diff
@@ -0,0 +1,37 @@
...
+    val registeredItems: Map<K, T>
+    fun register(item: T, key: K? = null)
+}
+
+/**
+ * A mutable implementation of [Registry] backed by a [MutableStateFlow].
+ * This abstract class allows runtime extension of the registry and optionally supports automatic
+ * key extraction from the item using the [extractKey] function.
+ */
+abstract class MutableRegistry<K, T>(
```
JesusMcCloud: ideally, this should be an interface with default implementations of functions. Of course, then the encapsulation with `registeredItems` wont work. I think we need to discuss this offline with @iaik-jheher, as he will be implementing more advanced registration and mapping, so for now, something simple and functional is fine, but we should probably make those interfaces internal anyway until we settle on something.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -91,35 +93,89 @@ enum class X509SignatureAlgorithm(
...
+        val PS256 = RSA(KnownOIDs.rsaPSS, "PS256", 256)
+        val PS384 = RSA(KnownOIDs.rsaPSS, "PS384", 384)
+        val PS512 = RSA(KnownOIDs.rsaPSS, "PS512", 512)
+
+        val RS1   = RSA(KnownOIDs.sha1WithRSAEncryption, "RS1")
+        val RS256 = RSA(KnownOIDs.sha256WithRSAEncryption, "RS256")
+        val RS384 = RSA(KnownOIDs.sha384WithRSAEncryption, "RS384")
+        val RS512 = RSA(KnownOIDs.sha512WithRSAEncryption, "RS512")
+
+        override val _registeredItems = MutableStateFlow(
```
JesusMcCloud: I really like how readable this is by overrding theregisterdItems, but this should probably somehow be encapsulated, so that we don't need to access a member that starts with an underscore. but also here: part of an offline discussion we need to have

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/Registry.kt
StjepanovicSrdjan: Is something like `transform` good enough or is it too generic, or should I use something that still indicates it is used to transform an item into a key (`keySelector`, `itemToKey`)?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -91,35 +94,78 @@ enum class X509SignatureAlgorithm(
...
+        val PS256 = RSA(KnownOIDs.rsaPSS, "PS256", 256)
+        val PS384 = RSA(KnownOIDs.rsaPSS, "PS384", 384)
+        val PS512 = RSA(KnownOIDs.rsaPSS, "PS512", 512)
+
+        val RS1   = RSA(KnownOIDs.sha1WithRSAEncryption, "RS1")
+        val RS256 = RSA(KnownOIDs.sha256WithRSAEncryption, "RS256")
+        val RS384 = RSA(KnownOIDs.sha384WithRSAEncryption, "RS384")
+        val RS512 = RSA(KnownOIDs.sha512WithRSAEncryption, "RS512")
+
+        val registeredAlgorithms = setOf(
```
JesusMcCloud: We called such constructs entries in other locations

```suggestion
        val entries = setOf(

File: indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt [hunk 13]
```diff
@@ -91,35 +94,78 @@ enum class X509SignatureAlgorithm(
...
+            ES256, ES384, ES512,
+            PS256, PS384, PS512,
+            RS1, RS256, RS384, RS512
+        )
+
+        private fun fromOid(oid: ObjectIdentifier): X509SignatureAlgorithm? =
+            registeredAlgorithms.firstOrNull { it.oid == oid }
```
JesusMcCloud:
```suggestion
            entries.firstOrNull { it.oid == oid }

File: indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt [hunk 14]
```diff
@@ -204,6 +250,6 @@ object X509SignatureAlgorithmSerializer : KSerializer<X509SignatureAlgorithm> {
 
     override fun deserialize(decoder: Decoder): X509SignatureAlgorithm {
         val decoded = decoder.decodeString()
-        return X509SignatureAlgorithm.entries.first { it.name == decoded }
+        return X509SignatureAlgorithm.registeredAlgorithms.first { it.name == decoded }
```
JesusMcCloud:
```suggestion
        return X509SignatureAlgorithm.entries.first { it.name == decoded }

File: indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/X509ConversionTests.kt [hunk 15]
```diff
@@ -11,7 +11,7 @@ infix fun <T> KmmResult<T>.shouldSucceedWith(b: T) : T =
     (this.getOrThrow() shouldBe b)
 class X509ConversionTests : FreeSpec({
     "X509 -> Alg -> X509 is stable" - {
-        withData(X509SignatureAlgorithm.entries) {
+        withData(X509SignatureAlgorithm.registeredAlgorithms) {
```
JesusMcCloud:
```suggestion
        withData(X509SignatureAlgorithm.entries) {

File: indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt (hunk 12 above)
Comment (StjepanovicSrdjan) [reply indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt]: Renamed in [ade6109](https://github.com/a-sit-plus/signum/pull/312/commits/ade610907f7350b65dd6b3c85fd6bf5a6222488a)

**Review by JesusMcCloud — APPROVED** [review]

File: gradle/libs.versions.toml [hunk 16]
```diff
@@ -20,6 +21,7 @@ core = { module = "androidx.test:core", version.ref = "core" }
 kotest-extensions-android = { module = "br.com.colman:kotest-extensions-android", version.ref = "kotestExtensionsAndroid" }
 kotest-runner-android = { module = "br.com.colman:kotest-runner-android", version.ref = "kotestRunnerAndroid" }
 kotlinx-io-core = { module = "org.jetbrains.kotlinx:kotlinx-io-core", version.ref = "kotlinxIoCore" }
+kotlinx-coroutines-core = { module = "org.jetbrains.kotlinx:kotlinx-coroutines-core", version.ref = "kotlinxCoroutinesCore" }
```
JesusMcCloud: remove

gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ ksp = "2.0.2"
 agp = "8.10.0"
 core = "1.5.0"
 kotlinxIoCore = "0.7.0"
+kotlinxCoroutinesCore = "1.10.2"
```
JesusMcCloud: remove

JesusMcCloud:
also add tests showcasing it actually works

gradle/libs.versions.toml
StjepanovicSrdjan: Fixed during rebase

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -91,35 +93,83 @@ enum class X509SignatureAlgorithm(
...
             ES512, PS512, RS512 -> Digest.SHA512
+            else -> throw IllegalArgumentException("Unsupported hash algorithm.")
         }
 
+    /**
+     * **Accessing this property may throw.**
+     * @throws IllegalArgumentException in case a non-supported [X509SignatureAlgorithm] is to be converted
+     */
+    // TODO update when core signature data classes become extensible
     override val algorithm: SignatureAlgorithm
```
JesusMcCloud: @iaik-jheher I really think we need to make the base property nullable to support arbitrary mappings. I will make another commit that pulls this through to get a feel for the consquences.

JesusMcCloud: @iaik-jheher  Do we also make `SpecializedSymmetricEncryptionAlgorithm`'s `algorithm` nullable while we're at it?

Issue #290, opened 2025-05-07
Is it necessary to add a method to try to decode the next child of `Asn1Structure`


ShiinaSekiu:
There are always many `OPTIONAL` tags in the asn1 structure, which makes it difficult to parse. Is it possible to add a method (maybe the method name can be better?)
`fun <T> Asn1Structure.tryDecodeNextChild(assertTag: Asn1Element.Tag, transform: (content: Asn1Element) -> T?) : T?`
The general logic is that if there is still a value and the value's tag is `assertTag`, then call `transform` and consume it. If not, then do not consume it and return `null`.

JesusMcCloud:
Just for clarification (thumbs-up/down reaction suffices) to make sure that I understand correctly:

* `nextChildOrNull()` will always consume any possibly present child, so it is not directly possible to only consume, if the child matches the asserted tag.
* Equivalent of the desired API using the currently available functionality could be something along the lines of the following:
```kotlin
someStructure.peek()?.let {//peek, but not consume next child
    //if the tag matches, consume
    if(it.tag == assertTag) src.nextChild().let { /*transfomration here*/ }
    //return null otherwise
    else null
}
```

@iaik-jheher any suggestions for the name of the proposed function? I'm not 100% happy, but lack better ideas myself.
Also I'd make it an inline fun and a `reified T` for more convenience

ShiinaSekiu:
Also I'd like to ask, is there any advantage to having him as an `inline fun` instead of putting it in `Asn1Structure`?

JesusMcCloud:
I don't fully understand. member functions can also be inline. it needs to be inline for `reified T` to work

ShiinaSekiu:
Sorry, I misunderstood, I was thinking that if you were inline, you wouldn't be able to get the `index` which is private, but if you use the `peek()` function to get it, you might not have such a problem.

ShiinaSekiu:
Also, do you think it's necessary to expose `index` reads and add `resetIndex()` `previousChild()`? This would greatly increase flexibility.
```
var index : Int = 0
    private set

fun resetIndex() { index = 0 }

@Throws(Asn1StructuralException::class)
fun previousChild() =
    catching { children[--index] }.getOrElse { throw Asn1StructuralException("No more content left") }
```

ShiinaSekiu:
```
EUICCInfo2 ::= [34] SEQUENCE { -- Tag 'BF22' 
[...]

PR #313, opened 2025-07-03
Enhancement/asn1 structure children decoding

StjepanovicSrdjan:
Add features suggested in #290 
I’m not completely sure about the `nextChildIfTagOrNull` function name — it seems closer to our current naming conventions than the proposed `tryDecodeNextChild`.


indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -498,8 +500,40 @@ sealed class Asn1Structure(
...
      */
-    fun peek() = if (!hasMoreChildren()) null else children[index]
+    fun peekForward() = if (!hasMoreChildren()) null else children[index]
+
+    /**
+     * Returns the previous child held by this structure
+     * @throws [Asn1StructuralException] if index points to the first child
+     */
+    @Throws(Asn1StructuralException::class)
+    fun previousChild() =
```
iaik-jheher: just commenting a gut reaction, but `previousChild` actually "rewinding" the children iterator feels weird

JesusMcCloud: @iaik-jheher @StjepanovicSrdjan @ShiinaSekiu  
Maybe we can go for something like the following (very half-baked idea, RFC from everyone welcome)
* let's deprecate the `nextChild` and `peek` functions and instead instead introduce a `childIterator()` function that, when called, creates a new iterator with a fresh index starting at zero.
* introduce a `reversed()` function on said iterator that reverses the order (but keeps the index where it currently is).
* tryDecode() as introduced here will be available on the iterator (naming discussion aside).

iaik-jheher: i agree with the first intuition and have had this same idea before (i do not like the iteration state being a property of the object itself)

i don't like `reversed()` operating in place (i am not sure if this was the intended suggestion?); it can return a new reverse iterator

JesusMcCloud: yes, sorry. not in place, but a fresh iterator that goes in reverse from the current index

StjepanovicSrdjan: I implemented the iterator, you can check out if I transferred the suggested idea properly. 

Now with the iterator in place, how should we deal with the trailing data check in doDecode.
Current implementation:
```@Throws(Asn1Exception::class)
    fun decodeFromTlv(src: A, assertTag: Asn1Element.Tag? = null, requireFullConsumption: Boolean = true): T {
        verifyTag(src, assertTag)
        return doDecode(src).also {
            if (requireFullConsumption && src is Asn1Structure) {
                runRethrowing {
                    require(!src.hasMoreChildren()) { "Trailing data found in ASN.1 structure" }
                }
            }
        }
    }```

**Review by JesusMcCloud — CHANGES_REQUESTED** [review]

File: indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt [hunk 2]
```diff
@@ -467,73 +467,65 @@ sealed class Asn1Structure(
...
-        catching { children[--index] }.getOrElse { throw Asn1StructuralException("No previous content left") }
+     * Returns a new iterator with a fresh index starting at 0
+     * */
+    fun childIterator(): ChildIterator = ChildIterator(children)
 
-    /**
-     * Exception-free version of [previousChild]
-     */
-    fun previousChildOrNull() = catchingUnwrapped { previousChild() }.getOrNull()
+    class ChildIterator(
```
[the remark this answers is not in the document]
JesusMcCloud: I am also unsure whether this class should actually implement `Iterator<Asn1Element` or not; after all, you can always access the child nodes which is a list, and therefore already a real `iterator`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: excellent question! I did not consider this. The iteration state must indeed be a property of the object itself. After all, we really want to enforce consuming everything, so an `Asn1Structure` really should be stateful when used inside a `doDecode`. Accessing children for getting a hold of them can be done using the `children` property at any point.

Maybe the answer is much simpler and the following is all you really need: 

```
var childIndex : Int = 0
    private set

fun <T> nextIfTagOrNull(assertTag: Tag, transform: (Asn1Element) -> T?): T? //still not happy about the name
```

This would keep everything working and behaving as expected, while providing all the flexibility if you really need intricate child node processing. Just getting hold of the child index at any time and the `nextIfTagOrNull` function introduced in this PR is probably enough to accomplish everything ever needed. After all, if you can read the index you can iterate over all children in reverse from where you are, if you really need it.

I am really not happy about the name `nextIfTagOrNull`, even though I don't yet have any better ideas.

StjepanovicSrdjan: Does the name `tryDecodeNextChild` suggested in the issue make more sense? Or I should leave it as is for now?

JesusMcCloud: naming things is hard. @iaik-jheher any other suggestions?

iaik-jheher: shouldn't `doDecode` just take an iterator?

JesusMcCloud: doDecode works on the current element you want to decide

iaik-jheher: cf 0a13e9af8dd1cbc24fa9d1e6914331ab56f8566f for a demonstrator of what that would look like

StjepanovicSrdjan: Suggested Iterator-based implementation finished in  [14374ed](https://github.com/a-sit-plus/signum/pull/313/commits/14374edc96dac4ecc865385e9804fdc65e1ca9f4), and [7e6e971](https://github.com/a-sit-plus/signum/pull/313/commits/7e6e9710813f23f7bdaf28c51c7078ec7b200a09)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
+         * Returns the current child or `null`, if there are no children left
+         * (useful when iterating over this structure's children).
+         */
+        fun peek() = if (!hasMoreChildren()) null else children[index]
+
+        /**
+         * Returns iterator with reversed view
+         * If [reverseIndex] is true, adjusts the starting index to point to the current element. Otherwise, copies the current index.
+         * */
+        fun reversed(reverseIndex: Boolean = false): Iterator =
```
JesusMcCloud: the index should always be reversed so that a reversed iterator will always be usable for rewinding from the current position

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Exception.kt
```diff
@@ -19,4 +19,12 @@ class Asn1OidException(message: String, val oid: ObjectIdentifier) : Asn1Excepti
  * Runs [block] inside [catching] and encapsulates any thrown exception in an [Asn1Exception] unless it already is one
  */
 @Throws(Asn1Exception::class)
-inline fun <reified R> runRethrowing(block: () -> R) = catchingUnwrappedAs(::Asn1Exception, block).getOrThrow()
\ No newline at end of file
+inline fun <reified R> runRethrowing(block: () -> R) = catchingUnwrappedAs(::Asn1Exception, block).getOrThrow()
+
+inline fun <reified R> Asn1Structure.decodeRethrowing(
```
JesusMcCloud, line 34: API docs should go here, as it is generally useful

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
      * (useful when iterating over this structure's children).
      */
-    fun peek() = if (!hasMoreChildren()) null else children[index]
+    @Deprecated(
+        message = "Use Iterator instead",
+        replaceWith = ReplaceWith("iterator().peek()"),
+        level = DeprecationLevel.ERROR
+    )
+    fun peek() = iterator.peek()
+
```
JesusMcCloud: I think these deserve some more explanation, that `iterator().whatever` will always produce a fresh iterator with index 0. To be safe, it might even be better to not add a `replaceWith` at all, but add a code snippet that shows how to properly use the new API to the API docs of all those deprecated functions

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
+
+    inner class Iterator(
```
JesusMcCloud: API docs missing for the class

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
+        private var index: Int = 0
+    ): kotlin.collections.Iterator<Asn1Element> {
+
+        /**
+         * Returns the next child held by this structure. Useful for iterating over its children when parsing complex structures.
+         * @throws [Asn1StructuralException] if no more children are available
+         */
+        @Throws(Asn1StructuralException::class)
+        fun nextChild() =
+            catching { children[index++] }.getOrElse { throw Asn1StructuralException("No more content left") }
```
JesusMcCloud: `runRethrowing` should work here instead of `catching`

JesusMcCloud: also since we're now implementing an interator, we should stick to the interface function names and sratch the custom named ones. move this function body to `next()`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
+
+        /**
+         * Exception-free version of [nextChild]
+         */
+        fun nextChildOrNull() = catchingUnwrapped { nextChild() }.getOrNull()
+
+        /**
+         * Returns `true` if more children can be retrieved by [nextChild]. `false` otherwise
+         */
+        fun hasMoreChildren() = children.size > index
```
JesusMcCloud: scratch that in favor of `hasNext()`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +474,100 @@ sealed class Asn1Structure(
...
+         * @throws [Asn1StructuralException] if no more children are available
+         */
+        @Throws(Asn1StructuralException::class)
+        fun nextChild() =
+            catching { children[index++] }.getOrElse { throw Asn1StructuralException("No more content left") }
+
+        /**
+         * Exception-free version of [nextChild]
+         */
+        fun nextChildOrNull() = catchingUnwrapped { nextChild() }.getOrNull()
```
JesusMcCloud: remove this to keep in line with iterator function naming conventions check if kotlin already provides an extension `nextOrNull` on all iterators. removing suffices, otherwise, rename to `nextOrNull`

JesusMcCloud (changes requested):
I'll review the changes that were needed throughout the codebase, once the iteraotr function renaming is done.

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
iaik-jheher: i believe `next()` is required to throw a standard exception (`NoSuchElementException`?)

JesusMcCloud: might well be. then we should too

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -518,10 +516,13 @@ sealed class Asn1Structure(
...
         level = DeprecationLevel.ERROR
     )
-    fun peek() = if (!hasMoreChildren()) null else children[index]
+    fun peek() = iterator.peek()
 
 
-    inner class Iterator: kotlin.collections.Iterator<Asn1Element> {
+    inner class Iterator(
+        private val children: List<Asn1Element>,
+        private var index: Int = 0
```
ShiinaSekiu: Should `index` be made externally accessible?
```
var index: Int = 0
    private set
```
or
```
private var _index: Int = 0
val index = _index
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
ShiinaSekiu: I think we can achieve this by providing a `direction` which is an Int and can only be 1 or -1, with 1 being positive stepping.
```
fun nextChild() =
        catching { 
            val current = children[index] }.getOrElse { throw Asn1StructuralException("No more content left")
            index += direction
            return current
        }
```
This ensures that the index does not change when reversing. Also, you may consider adding `val isForward = direction == 1` to provide the direction.
What do you think?

StjepanovicSrdjan: fixed in [de85a48](https://github.com/a-sit-plus/signum/pull/313/commits/de85a485dfbeec02fce8853251c58231a0ea802e)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: fixed in [de85a48](https://github.com/a-sit-plus/signum/pull/313/commits/de85a485dfbeec02fce8853251c58231a0ea802e) and [a079a50](https://github.com/a-sit-plus/signum/pull/313/commits/a079a501b9fb0ac15a8ef04d318eb172a63d6dda)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: fixed in [de85a48](https://github.com/a-sit-plus/signum/pull/313/commits/de85a485dfbeec02fce8853251c58231a0ea802e)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: fixed in [de85a48](https://github.com/a-sit-plus/signum/pull/313/commits/de85a485dfbeec02fce8853251c58231a0ea802e)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: Followed instructions in [a079a50](https://github.com/a-sit-plus/signum/pull/313/commits/a079a501b9fb0ac15a8ef04d318eb172a63d6dda). Not completely sure about current code snippets in API docs of deprecated functions. I covered both decoding(decodeRethrowing) and manual iterator cases

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Exception.kt
StjepanovicSrdjan: fixed in [a079a50](https://github.com/a-sit-plus/signum/pull/313/commits/a079a501b9fb0ac15a8ef04d318eb172a63d6dda)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -468,61 +468,140 @@ sealed class Asn1Structure(
...
     ): kotlin.collections.Iterator<Asn1Element> {
 
         /**
          * Returns the next child held by this structure. Useful for iterating over its children when parsing complex structures.
-         * @throws [Asn1StructuralException] if no more children are available
+         * @throws [NoSuchElementException] if no more children are available
          */
         override fun next() =
-            runRethrowing { children[index++] }
+            catching { children[index++] }.getOrElse { throw NoSuchElementException("No more content left") }
```
JesusMcCloud: catchingUnwrapped, since we throw directly

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: definitely the first option. but yes, it should be accessible

ShiinaSekiu:
@JesusMcCloud I think we can implement iterators by providing a `direction`, which is an Int type that can only be 1 or -1, where 1 means moving in the forward direction.
Example:
```
fun nextChild() =
        catching { 
            val current = children[index] }.getOrElse { throw Asn1StructuralException("No more content left")
            index += direction
            return current
        }
```
This ensures that the index does not change when reversing. Also, you may consider adding `val isForward = direction == 1` to get the direction. (Of course I'm not sure about the name)
What do you think?

JesusMcCloud:
> @JesusMcCloud I think we can implement iterators by providing a `direction`, which is an Int type that can only be 1 or -1, where 1 means moving in the forward direction. Example:
> 
> ```
> fun nextChild() =
>         catching { 
>             val current = children[index] }.getOrElse { throw Asn1StructuralException("No more content left")
>             index += direction
>             return current
>         }
> ```
> 
> This ensures that the index does not change when reversing. Also, you may consider adding `val isForward = direction == 1` to get the direction. (Of course I'm not sure about the name) What do you think?

Exposing a direction could be beneficial, but we should really stick with the iterator interface function names and not use custom names for functionality as defined by the interface.

ShiinaSekiu:
> > @JesusMcCloud I think we can implement iterators by providing a `direction`, which is an Int type that can only be 1 or -1, where 1 means moving in the forward direction. Example:
> > ```
> > fun nextChild() =
> >         catching { 
> >             val current = children[index] }.getOrElse { throw Asn1StructuralException("No more content left")
> >             index += direction
> >             return current
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This ensures that the index does not change when reversing. Also, you may consider adding `val isForward = direction == 1` to get the direction. (Of course I'm not sure about the name) What do you think?
> 
> Exposing a direction could be beneficial, but we should really stick with the iterator interface function names and not use custom names for functionality as defined by the interface.

I don't really care about the direction, my focus is actually on the index, which is now reversed. The index obtained cannot be used on the original `Asn1Structure` (which is also reversed)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
      */
     val isActuallySorted: Boolean by if (sortChildren) lazyOf(true) else lazy { children.sortedBy { it.tag } == children }
 
-    private var index = 0
+    private val iterator by lazy { iterator() }
```
iaik-jheher: can we call this `backwardsCompatibilityIterator` or something

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
 
     /**
      * Returns the next child held by this structure. Useful for iterating over its children when parsing complex structures.
      * @throws [Asn1StructuralException] if no more children are available
+     *### Migration Examples:
+     *
+     * Using `decodeRethrowing`:
+     * ```kotlin
+     * val result = structure.decodeRethrowing {
```
iaik-jheher, line 537: do we expect consumers to use `decodeRethrowing`?

`runRethrowing`/`decodeRethrowing` are internal-use things that convert exceptions, right?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
+     * ```kotlin
+     * val iterator = structure.iterator()
+     * while (iterator.hasNext()) {
+     *     val child = iterator.next()
+     *     // Process child
+     * }
+     * ```
      */
+    @Deprecated(
+        message = "Use Iterator inside decodeAs/decodeRethrowing",
```
iaik-jheher:
```suggestion
        message = "Use an explicit iterator()",
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
+     * val iterator = structure.iterator()
+     * while (iterator.hasNext()) {
+     *     val child = iterator.nextOrNull()
+     *     // Process child
+     * }
+     * ```
      */
-    fun nextChildOrNull() = catchingUnwrapped { nextChild() }.getOrNull()
+    @Deprecated(
+        message = "Use Iterator inside decodeAs/decodeRethrowing",
```
iaik-jheher: here too

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
+     * ```kotlin
+     * val iterator = structure.iterator()
+     * if (iterator.hasNext()) {
+     *     // ...
+     * }
+     * ```
      */
-    fun hasMoreChildren() = children.size > index
+    @Deprecated(
+        message = "Use Iterator inside decodeAs/decodeRethrowing",
```
iaik-jheher: here too

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
+     * ```kotlin
+     * val iterator = structure.iterator()
+     * val child = iterator.peek()
+     * if (child != null) {
+     *     // Inspect child without consuming it
+     * }
+     * ```
+     */
+    @Deprecated(
+        message = "Use Iterator inside decodeAs/decodeRethrowing",
```
iaik-jheher: here too

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +467,180 @@ sealed class Asn1Structure(
...
+        fun peek() = if (!hasNext()) null else children[index]
+
+        /**
+         * Returns iterator with reversed view
+         * */
+        fun reversed(): Iterator =
+            Iterator(children.reversed(), children.lastIndex - index)
+    }
+
+    override operator fun iterator() = Iterator(children)
```
iaik-jheher: why are we passing `children` into `Iterator`'s constructor if `Iterator` is an inner class anyway?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Exception.kt
```diff
@@ -19,4 +19,18 @@ class Asn1OidException(message: String, val oid: ObjectIdentifier) : Asn1Excepti
...
\ No newline at end of file
+inline fun <reified R> runRethrowing(block: () -> R) = catchingUnwrappedAs(::Asn1Exception, block).getOrThrow()
+
+/**
+ * Decodes this ASN.1 structure using the provided [decoder] lambda, rethrowing any caught exception
+ * as an [Asn1Exception].
+ * This is a wrapper around [Asn1Structure.decodeAs] that ensures exceptions thrown during decoding are
+ * consistently rethrown as [Asn1Exception], using the [runRethrowing] utility.
+ */
+inline fun <reified R> Asn1Structure.decodeRethrowing(
```
iaik-jheher, line 34: shouldn't this be `internal`? or perhaps moving it to the internals module

iaik-jheher (commented):
structural review, i will look at the various dependent refactors later

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Exception.kt
JesusMcCloud: i think not. if you are going to define your own external asn.1 structures, this is really handy

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: so we can keep the iteration logic exactly the same for reversed iterators. simpler overall code at the cost of a single reference

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
JesusMcCloud: decodeRethrowing is handy externally as well

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: Fixed in [9eb37aa](https://github.com/a-sit-plus/signum/pull/313/commits/9eb37aa5fa87a6ce8af072c3aa82b51c23a5fb51)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: Fixed in [9eb37aa](https://github.com/a-sit-plus/signum/pull/313/commits/9eb37aa5fa87a6ce8af072c3aa82b51c23a5fb51)

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: With the reversing using direction property as suggested in https://github.com/a-sit-plus/signum/pull/313#issuecomment-3060735769 `Iterator`'s constructor is no longer accepting children.

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
StjepanovicSrdjan: fixed in [0982c9b](https://github.com/a-sit-plus/signum/pull/313/commits/0982c9b98133a254252d4499c0050d03e42fa9b6) and [9eb37aa](https://github.com/a-sit-plus/signum/pull/313/commits/9eb37aa5fa87a6ce8af072c3aa82b51c23a5fb51)

docs/docs/indispensable-asn1.md
```diff
@@ -99,6 +99,11 @@ Hence, a typical decoding pipeline looks as follows:
 
 #### High-Level
 
+!!! Tip
+    Structured, iterator-based decoding of `Asn1Structure` has been introduced.
+    The functions `nextChild()`, `nextChildOrNull()`, `hasMoreChildren()`, and `peek()` are now deprecated in favor of the new iterator approach.
+    Child access is now performed using the inner Iterator, which provides the following methods: `next()`, `nextOrNull()`, `hasNext()`, `peek()`, and `reversed()`.
```
JesusMcCloud: I'd concretely mention that `Asn1Structure` now implements `Iterable<Asn1Element>` and thus provides the `iterator()` function to get a child iterator. also: inner **i**terator (lowercase)

JesusMcCloud (approved):
minor api doc nitpick

docs/docs/indispensable-asn1.md
iaik-jheher:
```suggestion
    Structured, iterator-based decoding of `Asn1Structure` has been introduced.
    The functions `nextChild()`, `nextChildOrNull()`, `hasMoreChildren()`, and `peek()` are now deprecated in favor of the new iterator approach.
    Child access is now performed using the inner `iterator()`, which provides the following methods: `next()`, `nextOrNull()`, `hasNext()`, `peek()`, and `reversed()`.
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Encodable.kt
```diff
@@ -82,13 +82,7 @@ interface Asn1Decodable<A : Asn1Element, out T : Asn1Encodable<A>> {
     @Throws(Asn1Exception::class)
     fun decodeFromTlv(src: A, assertTag: Asn1Element.Tag? = null, requireFullConsumption: Boolean = true): T {
```
iaik-jheher: remove boolean arg that does nothing

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -463,26 +462,27 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
                 require(version == 1) { "EC public key version must be 1" }
-                val privateKeyOctets = src.nextChild().asOctetString().content
+                val privateKeyOctets = next().asOctetString().content
 
                 var curve: ECCurve? = null
                 var publicKey: Asn1BitString? = null
-                while (src.hasMoreChildren()) {
-                    val elm= src.nextChild().asExplicitlyTagged()
+                while (hasNext()) {
+                    val elm= next().asExplicitlyTagged()
```
iaik-jheher:
```suggestion
                    val elm = next().asExplicitlyTagged()
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -463,26 +462,27 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
+                    val elmIterator = elm.iterator()
                     when (elm.tag.tagValue) {
                         0uL -> {
                             require(curve == null) { "Duplicate EC curve field in EC PrivateKey" }
                             require(publicKey == null) { "Field order violation in EC PrivateKey" }
                             require(elm.children.size == 1) { "Invalid EC curve field in EC PrivateKey" }
-                            curve = ObjectIdentifier.decodeFromTlv(elm.nextChild().asPrimitive()).let(ECCurve::withOid)
+                            curve = ObjectIdentifier.decodeFromTlv(elmIterator.next().asPrimitive()).let(ECCurve::withOid)
```
iaik-jheher: if we are explicitly checking `elm.children.size` above then `elm.children[0]` feels correct here

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -463,26 +462,27 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
                         }
                         1uL -> {
                             require(publicKey == null) { "Duplicate public key field in EC PrivateKey" }
                             require(elm.children.size == 1) { "Invalid public key field in EC PrivateKey" }
-                            publicKey = elm.nextChild().asPrimitive().asAsn1BitString()
+                            publicKey = elmIterator.next().asPrimitive().asAsn1BitString()
```
iaik-jheher: as above

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -493,9 +493,11 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
                     attributes,
                     privateKeyOctets.size)
-                return@runRethrowing when(curve) {
-                    null -> privateKey
-                    else -> privateKey.withCurve(curve)
+                runRethrowing {
+                    when (curve) {
+                        null -> privateKey
+                        else -> privateKey.withCurve(curve)
+                    }
```
iaik-jheher: unclear what this change is meant to do?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -521,21 +523,21 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
-            require(!algorithmID.hasMoreChildren()) { "Superfluous Algorithm ID data encountered" }
-            val privateKeyStructure = src.nextChild().asEncapsulatingOctetString().let {
-                val seq = it.nextChild().asSequence()
-                require(!it.hasMoreChildren()) { "Superfluous private key data encountered" }
+        override fun doDecode(src: Asn1Sequence): CryptoPrivateKey = src.decodeRethrowing {
+            require(next().asPrimitive().decodeToInt() == 0) { "PKCS#8 Private Key VERSION must be 0" }
+            val algorithmIDIterator = next().asSequence().iterator()
+            val algIdentifier = ObjectIdentifier.decodeFromTlv(algorithmIDIterator.next().asPrimitive())
+            val algParams = algorithmIDIterator.next().asPrimitive()
+            require(!algorithmIDIterator.hasNext()) { "Superfluous Algorithm ID data encountered" }
```
iaik-jheher: shouldn't this be `next().asSequence().decodeRethrowing { ... }`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -521,21 +523,21 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
+            val privateKeyStructure = next().asEncapsulatingOctetString().iterator().let {
+                val seq = it.next().asSequence()
+                require(!it.hasNext()) { "Superfluous private key data encountered" }
                 seq
             }
```
iaik-jheher: `decodeRethrowing` which gives you the superfluous data check for free?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -108,34 +106,35 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
 
 
         @Throws(Asn1Exception::class)
-        override fun doDecode(src: Asn1Sequence): CryptoPublicKey = runRethrowing {
+        override fun doDecode(src: Asn1Sequence): CryptoPublicKey = src.decodeRethrowing {
             if (src.children.size != 2) throw Asn1StructuralException("Invalid SPKI Structure!")
-            val keyInfo = src.nextChild() as Asn1Sequence
+            val keyInfo = next() as Asn1Sequence
+            val keyInfoIterator = keyInfo.iterator()
             if (keyInfo.children.size != 2) throw Asn1StructuralException("Superfluous data in  SPKI!")
```
iaik-jheher, line 112: here too, IMO; if you specifically check `children.size` then access `children` directly instead of mixing with iterators

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -108,34 +106,35 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
...
-                    val n = (rsaSequence.nextChild() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
-                    val e = (rsaSequence.nextChild() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
-                    if (rsaSequence.hasMoreChildren()) throw Asn1StructuralException("Superfluous data in SPKI!")
-                    return RSA(n, e)
+                    (keyInfoIterator.next() as Asn1Primitive).readNull()
+                    val bitString = (next() as Asn1Primitive).asAsn1BitString()
+                    val rsaSequenceIterator = Asn1Element.parse(bitString.rawBytes).asSequence().iterator()
+                    val n = (rsaSequenceIterator.next() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
+                    val e = (rsaSequenceIterator.next() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
+                    if (rsaSequenceIterator.hasNext()) throw Asn1StructuralException("Superfluous data in SPKI!")
```
iaik-jheher: could be `decodeRethrowing`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -124,30 +124,30 @@ enum class X509SignatureAlgorithm(
         }
 
         @Throws(Asn1Exception::class)
-        private fun parsePssParams(src: Asn1Sequence): X509SignatureAlgorithm = runRethrowing {
-            val seq = src.nextChild() as Asn1Sequence
-            val first = (seq.nextChild() as Asn1ExplicitlyTagged).verifyTag(0u).single() as Asn1Sequence
+        private fun parsePssParams(src: Asn1Sequence): X509SignatureAlgorithm = src.decodeRethrowing {
```
iaik-jheher: review this and use nested `decodeRethrowing` where possible instead of this iterator variable mess

(i'm sure you will have missed the trailing data check for at least one of them)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/Pkcs10CertificationRequest.kt
```diff
@@ -84,18 +82,18 @@ data class TbsCertificationRequest(
...
-            val attributes = if (src.hasMoreChildren()) {
-                (src.nextChild() as Asn1ExplicitlyTagged).verifyTag(0u)
+            val cryptoPublicKey = CryptoPublicKey.decodeFromTlv(next() as Asn1Sequence)
+            val attributes = if (hasNext()) {
+                (next() as Asn1ExplicitlyTagged).verifyTag(0u)
                     .map { Pkcs10CertificationRequestAttribute.decodeFromTlv(it as Asn1Sequence) }
             } else null
 
-            if (src.hasMoreChildren()) throw Asn1StructuralException("Superfluous Data in CSR Structure")
+            if (hasNext()) throw Asn1StructuralException("Superfluous Data in CSR Structure")
```
iaik-jheher: superfluous, done by `decodeRethrowing`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/Pkcs10CertificationRequest.kt
```diff
@@ -155,12 +153,12 @@ data class Pkcs10CertificationRequest(
...
-            val tbsCsr = TbsCertificationRequest.decodeFromTlv(src.nextChild() as Asn1Sequence)
-            val sigAlg = X509SignatureAlgorithm.decodeFromTlv(src.nextChild() as Asn1Sequence)
-            val signature = (src.nextChild() as Asn1Primitive).asAsn1BitString()
-            if (src.hasMoreChildren()) throw Asn1StructuralException("Superfluous structure in CSR Structure")
-            return Pkcs10CertificationRequest(tbsCsr, sigAlg, CryptoSignature.decodeFromDer(signature.rawBytes))
+        override fun doDecode(src: Asn1Sequence): Pkcs10CertificationRequest = src.decodeRethrowing {
+            val tbsCsr = TbsCertificationRequest.decodeFromTlv(next() as Asn1Sequence)
+            val sigAlg = X509SignatureAlgorithm.decodeFromTlv(next() as Asn1Sequence)
+            val signature = (next() as Asn1Primitive).asAsn1BitString()
+            if (hasNext()) throw Asn1StructuralException("Superfluous structure in CSR Structure")
```
iaik-jheher, line 157: superfluous

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -114,13 +112,13 @@ sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
...
-            val oid = (src.nextChild() as Asn1Primitive).readOid()
+        override fun doDecode(src: Asn1Sequence): AttributeTypeAndValue = src.decodeRethrowing {
+            val oid = next().asPrimitive().readOid()
             if (oid.nodes.size >= 3 && oid.toString().startsWith("2.5.4.")) {
-                val asn1String = src.nextChild() as Asn1Primitive
+                val asn1String = next().asPrimitive()
                 val str = catching { (asn1String).asAsn1String() }
-                if (src.hasMoreChildren()) throw Asn1StructuralException("Superfluous elements in RDN")
-                return when (oid) {
+                if (hasNext()) throw Asn1StructuralException("Superfluous elements in RDN")
```
iaik-jheher: superfluous

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
JesusMcCloud: 1. there's no `runRethrowing` any more
2. this is now the toplevel last statement since switching to `decodeRethrowing` so no `return` any more

CHANGELOG.md
```diff
@@ -23,23 +32,23 @@
 * Clean up some function signatures:
     * `SymmetricKey.toJsonWebKey` now returns `KmmResult`
     * `SymmetricEncryptionAlgorithm.toJweKwAlgorithm` now returns `KmmResult`
-    * `SymmetricEncryptionAlgorithm.toJweEncryptionAlgorithm` removed* Dependency Updates:
-* `kotlincrypto:secure-random:0.3.2` -> `kotlincrypto.random:crypto-rand:0.5.0`
-    * This fixes key generation in WASM/JS
-* Deprecate `serialize()` and `deserialize()` methods in data classes
```
iaik-jheher, line 37: i think the serialize/deserialize deprecation was accidentally removed when rejiggling this file

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +466,206 @@ sealed class Asn1Structure(
...
+        override fun next() =
+            catchingUnwrapped {
+                currentIndex = nextIndex
+                currentElement
+            }.getOrElse { throw NoSuchElementException("No more content left") }
+
+        /**
+         * Exception-free version of [next]
+         */
+        fun nextOrNull() = catchingUnwrapped { next() }.getOrNull()
```
iaik-jheher:
```suggestion
        fun nextOrNull() = if (hasNext()) next() else null
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +466,206 @@ sealed class Asn1Structure(
...
+
+        /**
+         * Returns the next child held by this structure. Useful for iterating over its children when parsing complex structures.
+         * @throws [NoSuchElementException] if no more children are available
+         */
+        override fun next() =
+            catchingUnwrapped {
+                currentIndex = nextIndex
+                currentElement
+            }.getOrElse { throw NoSuchElementException("No more content left") }
```
iaik-jheher:
```suggestion
        override fun next() {
            currentIndex = nextIndex
            return currentElement
        }
```

`currentElement` will already throw `NoSuchElementException`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Elements.kt
```diff
@@ -474,31 +466,206 @@ sealed class Asn1Structure(
...
+    }
+
+    /**
+     * Decodes the content of this ASN.1 structure using the provided [decoder] lambda.
+     * This function gives a convenient way to decode ASN.1 structures by exposing an
+     * iterator over the structure's children to the [decoder] lambda. Optionally, it enforces that
+     * all children must be consumed. Use [decodeRethrowing] to automatically and consistently wrap exceptions
+     * thrown during decoding in [Asn1Exception]s.
+     */
+    fun <T> decodeAs(requireFullConsumption: Boolean = true, decoder: Iterator.() -> T): T {
```
iaik-jheher:
```suggestion
    inline fun <reified T> decodeAs(requireFullConsumption: Boolean = true, decoder: Iterator.() -> T): T {
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Exception.kt
```diff
@@ -19,4 +19,18 @@ class Asn1OidException(message: String, val oid: ObjectIdentifier) : Asn1Excepti
...
+    requireFullConsumption: Boolean = true,
+    noinline decoder: Asn1Structure.Iterator.() -> R
```
iaik-jheher:
```suggestion
    decoder: Asn1Structure.Iterator.() -> R
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -145,14 +146,14 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
-                require(product.times(info.coefficient).mod(info.prime) == BigInteger.ONE)
-                    { "t_$i != (r_0 * ... * r_${i-1})^(-1) mod r_$i" }
-                product *= info.prime
-                require(info.exponent == privateKey.mod(pminusone)) { "d_$i != d mod (p_$i - 1)" }
-                require(e.multiply(info.exponent).mod(pminusone) == BigInteger.ONE)
-            }
+                .forEachIndexed { i, info ->
+                    val pminusone = info.prime - BigInteger.ONE
+                    require(product.times(info.coefficient).mod(info.prime) == BigInteger.ONE)
+                    { "t_$i != (r_0 * ... * r_${i - 1})^(-1) mod r_$i" }
```
iaik-jheher:
```suggestion
                    { "t_$i != (r_0 * ... * r_${i-1})^(-1) mod r_$i" }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -492,10 +505,13 @@ sealed interface CryptoPrivateKey : PemEncodable<Asn1Sequence>, Identifiable {
...
-                    null -> privateKey
-                    else -> privateKey.withCurve(curve)
+                    privateKeyOctets.size
+                )
+                runRethrowing {
+                    when (curve) {
+                        null -> privateKey
+                        else -> privateKey.withCurve(curve)
+                    }
                 }
```
iaik-jheher:
```suggestion
                when (curve) {
                    null -> privateKey
                    else -> privateKey.withCurve(curve)
                }
```
`decodeRethrowing already encapsulates exceptions`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -234,11 +236,12 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
              */
             @Throws(Asn1Exception::class)
             fun fromPKCS1encoded(input: ByteArray): RSA = runRethrowing {
-                val conv = Asn1Element.parse(input) as Asn1Sequence
-                val n = (conv.nextChild() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
-                val e = (conv.nextChild() as Asn1Primitive).decodeToAsn1Integer() as Asn1Integer.Positive
-                if (conv.hasMoreChildren()) throw Asn1StructuralException("Superfluous bytes")
-                return RSA(n, e)
+                val conv = Asn1Element.parse(input).asSequence()
+                conv.decodeRethrowing {
```
iaik-jheher:
```suggestion
                Asn1Element.parse(input).asSequence().decodeRethrowing {
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
```diff
@@ -203,11 +203,12 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
             }
 
             override fun doDecode(src: Asn1Element): EC.IndefiniteLength {
```
iaik-jheher:
```suggestion
            override fun doDecode(src: Asn1Element) = decodeRethrowing {
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
```diff
@@ -203,11 +203,12 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
...
-                src as Asn1Sequence
-                val r = (src.nextChild() as Asn1Primitive).decodeToBigInteger()
-                val s = (src.nextChild() as Asn1Primitive).decodeToBigInteger()
-                if (src.hasMoreChildren()) throw Asn1Exception("Illegal Signature Format")
-                return fromRS(r, s)
+                return src.asSequence().decodeRethrowing {
```
iaik-jheher:
```suggestion
                src.asSequence().decodeRethrowing {
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/X509SignatureAlgorithm.kt
```diff
@@ -108,46 +108,50 @@ enum class X509SignatureAlgorithm(
...
-            if (tag != Asn1Element.Tag.NULL)
-                throw Asn1TagMismatchException(Asn1Element.Tag.NULL, tag, "PSS Params not supported yet")
-
-            val second = (seq.nextChild() as Asn1ExplicitlyTagged).verifyTag(1u).single() as Asn1Sequence
-            val mgf = (second.nextChild() as Asn1Primitive).readOid()
-            if (mgf != KnownOIDs.pkcs1_MGF) throw IllegalArgumentException("Illegal OID: $mgf")
-            val inner = second.nextChild() as Asn1Sequence
-            val innerHash = (inner.nextChild() as Asn1Primitive).readOid()
-            if (innerHash != sigAlg) throw IllegalArgumentException("HashFunction mismatch! Expected: $sigAlg, is: $innerHash")
+        private fun parsePssParams(src: Asn1Structure.Iterator): X509SignatureAlgorithm = runRethrowing{
```
iaik-jheher, line 127: and so much more readable too - nice!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -151,48 +151,48 @@ constructor(
...
+            val issuerUniqueID = peek()?.let { next ->
                 if (next.tag == ISSUER_UID) {
-                    (src.nextChild() as Asn1Primitive).let { Asn1BitString.decodeFromTlv(it, ISSUER_UID) }
+                    next()
+                        .asPrimitive().let { Asn1BitString.decodeFromTlv(it, ISSUER_UID) }
                 } else null
             }
 
-            val subjectUniqueID = src.peek()?.let { next ->
+            val subjectUniqueID = peek()?.let { next ->
```
iaik-jheher, line 183:
```kotlin
val subjectUniqueID = nextOrNull()?.takeIf { it.tag == SUBJECT_UID }?.let { Asn1BitString.decodeFromTlv(it) }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -151,48 +151,48 @@ constructor(
...
+            val timestamps = decodeTimestamps(next().asSequence())
+            val subject = (next().asSequence()).children.map {
+                RelativeDistinguishedName.decodeFromTlv(it.asSet())
             }
 
-            val cryptoPublicKey = CryptoPublicKey.decodeFromTlv(src.nextChild() as Asn1Sequence)
+            val cryptoPublicKey = CryptoPublicKey.decodeFromTlv(next().asSequence())
 
-            val issuerUniqueID = src.peek()?.let { next ->
+            val issuerUniqueID = peek()?.let { next ->
```
iaik-jheher, line 176:
```kotlin
val issuerUniqueID = nextOrNull()?.takeIf { it.tag == ISSUER_UID }?.let { Asn1BitString.decodeFromTlv(it) }
```

CHANGELOG.md
JesusMcCloud: actually, it was in there twice and i removed the duplicate

PR #339, opened 2025-08-20
Enhancement/cose mac0 support

StjepanovicSrdjan:
COSE_Mac0 support
Renamed and refactored several underlying classes from Cose_Sign1 for reuse in Cose_Mac0. Still a lot of boilerplate code remains, with plenty of room for improvement.


nodh (approved):
I think that's already in very good shape!

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
```diff
@@ -12,34 +12,46 @@ import kotlinx.serialization.decodeFromByteArray
...
+        externalAad: ByteArray = byteArrayOf(),
+        detachedPayload: ByteArray? = null,
+    ): ByteArray = toCoseInput(externalAad, detachedPayload, "MAC0")
+
     /**
      * @param detachedPayload only to be set when [payload] is null, i.e. it is transported externally,
      * as it ignores the [payload] member
      */
-    internal fun toCoseSignatureInput(
+    private fun toCoseInput(
```
JesusMcCloud, line 49: I'm not too versed in COSE, but can it make sense to make this `public` for custom contexts?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
```diff
@@ -12,34 +12,46 @@ import kotlinx.serialization.decodeFromByteArray
...
         externalAad: ByteArray = byteArrayOf(),
         detachedPayload: ByteArray? = null,
```
JesusMcCloud, line 51: I know it's been like that before, but while we're breaking stuff: why is one an empty array and another one null? I assume it's got something to do with the fact that signing an empty array is a legit use case but a null AAD is cryptographically treated the same as an empty array, so the empty array default value makes sense. Irrespective of making the function `public` some API docs explaining why those params default to different things would be nice.

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseMac.kt
```diff
@@ -0,0 +1,121 @@
+package at.asitplus.signum.indispensable.cosef
```
JesusMcCloud, line 1: I'm trusting that this is conforming to the spec

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
```diff
@@ -0,0 +1,141 @@
+package at.asitplus.signum.indispensable.cosef
```
JesusMcCloud, line 1: Those two seem very much the same. Checkout [TemplateSerializer](https://github.com/a-sit-plus/signum/blob/main/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt#L37) to deduplicate code where possible

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Serializes [CoseSigned] with a typed payload, by using its [CoseSigned.wireFormat].
+ * Also handles deserialization of the bytes.
+ */
+class CoseSignedSerializer<P : Any?>(
+    private val parameterSerializer: KSerializer<P>,
+) : KSerializer<CoseSigned<P>> {
+
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("CoseSigned", PrimitiveKind.BYTE)
```
JesusMcCloud: Not a single byte. I know the old one did that too, but it's not right. I don't really know how to do it properly in KxS, maybe go for string instead of byte?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
nodh: By [spec](https://www.rfc-editor.org/rfc/rfc9052.html#section-4.4-2.4) the default for `aad` is an empty array.

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
nodh: It should probably be `buildClassSerialDescriptor` like in [vc-k](https://github.com/a-sit-plus/vck/blob/83e825fc064f1efdfc4537033bf476fd8d70f775/openid-data-classes/src/commonMain/kotlin/at/asitplus/iso/IssuerSignedItemSerializer.kt#L23)?

JesusMcCloud: There is also [this](https://github.com/Kotlin/kotlinx.serialization/blob/13a487c8d14a106f1a55b4b96419a5e89918fceb/core/commonMain/src/kotlinx/serialization/builtins/BuiltinSerializers.kt#L194), but i don't think that works out for something "flat" (as opposed to a structure/container) of primitives

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
JesusMcCloud: if it is even in the spec, its fine

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
StjepanovicSrdjan: That is something internally used by `CoseMac` and `CoseSign`. @nodh what is your opinion since that was something implemented before?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseMac.kt
StjepanovicSrdjan: Since this is only Cose_Mac0 support, the Cose_Recipient isn't necessary

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
StjepanovicSrdjan: fixed in https://github.com/a-sit-plus/signum/pull/339/commits/eff1a9a98fce5ba790bb68a2c0a0758182c11414

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
StjepanovicSrdjan: Does https://github.com/a-sit-plus/signum/pull/339/commits/ff88be20bafa971be63cb390ac91f991c8b1dd80 fix the issue?

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseBytes.kt
nodh: I don't think we'll need this public, as the context strings are fixed by the RFC

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
JesusMcCloud: much better!

JesusMcCloud (approved):

indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseSerializer.kt
```diff
@@ -127,7 +125,7 @@ private fun <P : Any?> ByteArray.fromByteStringWrapper(serializer: KSerializer<P
         this
     ).value
 
-fun ByteArray.toHeader(): CoseHeader =
+fun ByteArray.toProtectedHeader(): CoseHeader =
```
StjepanovicSrdjan: Encoding protected header was already done correctly in previous implementation, only bug was in this decoding function where we missed check if it is 0-lenght byte, because if that's the case than we try to deserialize it using CoseHeaderSerializer which expects a map, so I just added that check and renamed the function to be more clear that it is only used for protected header. I considered creating separate serializers for protected and unprotected headers, but in the end this previous implementation seems pretty well overall with just two extension functions on top of the CoseHeaderSerializer. @nodh @JesusMcCloud thoughts?

JesusMcCloud: functionally, this is fine, but the pollution of bytearray with this functionality is not. I wonder, if an empty protected header is discernable from an unprotected header by virtue of the cose tag, why not

- move the `if` inside the coseheaderserializer
- peek the first byte (we have that for cosekey: you cast the decoder and then get the current byte)
- if it is an empty array, consume it, and then take the shorthand
- otherwise, do as before

am I missing something?

StjepanovicSrdjan: I introduced ProtectedCoseHeaderSerializer in https://github.com/a-sit-plus/signum/pull/339/commits/a3eb4365efa7ecdc5958c35a835aaa3cef4fa517 that deals with the serialization of protected header. Also removed storing protected header as raw bytes with calling ByteArray extension functions and replaced it with CoseHeader that uses new serializer. 
Slightly different approach then @JesusMcCloud suggested but it felt more natural

JesusMcCloud: https://github.com/a-sit-plus/signum/pull/339/commits/a3eb4365efa7ecdc5958c35a835aaa3cef4fa517 looks pretty nice! very straight-forward end comprehensible. if everything's green after the pipeline is through, merge it!

PR #345, opened 2025-08-28
Fix/asn1 string improvement

StjepanovicSrdjan:
- Deprecated `Asn1Primitive.asAsn1String()`, moved decoding logic into `Asn1String.doDecode()`
-  Added specific decodeTo functions in `Asn1Primitive` for every string type that follow usual pattern for decoding primitive, when dealing with implicit tags caller should call function for decoding specific string type. If we do something like 'Asn1String.decodeFromTlv()` we expect explicit tag.
- Added `Asn1String.rawValue` property for storing raw ByteArray of the string, used in checks for several charsets and good for reencoding Asn1String, it will be encoded as it was before decoding

- Possible way of enhancing this: we could use https://github.com/fleeksoft/fleeksoft-io/blob/main/CharsetsReadme.md library for charset support in KMP and decode in different charsets for different string types. Currently we relay on `ByteArray.decodeToString()` which decodes a string from the bytes in UTF-8 encoding


indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -328,25 +329,76 @@ inline fun Asn1Primitive.decodeToFloatOrNull(assertTag: Asn1Element.Tag = Asn1El
...
+        T61_STRING.toULong() -> Asn1String.Teletex(content)
+        PRINTABLE_STRING.toULong() -> Asn1String.Printable(content)
+        NUMERIC_STRING.toULong() -> Asn1String.Numeric(content)
+        VISIBLE_STRING.toULong() -> Asn1String.Visible(content)
         else -> TODO("Support other string tag $tag")
     }
 }
 
+@Throws(Asn1Exception::class)
+inline fun Asn1Primitive.decodeToUtf8String(assertTag: Asn1Element.Tag = Asn1Element.Tag.REAL) =
```
JesusMcCloud: copy-paste error

JesusMcCloud (commented):
I would not introduce any charsets except UTF-8 (unless I am missing something here) as this is beyond ASN.1. If someone needs that -> the `rawValue` you introduced can be used. Will do a full review once marked as ready

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,193 @@
...
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToPrintableString
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToTeletextString
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToUniversalString
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToUnrestrictedString
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToUtf8String
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToVideotexString
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToVisibleString
+
 
 //TODO auto-sanitize and/or reduce
```
JesusMcCloud: I think this can go

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,193 @@
...
 /**
  * ASN.! String class used as wrapper do discriminate between different ASN.1 string types
```
JesusMcCloud: my bang typo can be fixed too

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,193 @@
...
  */
-sealed class Asn1String : Asn1Encodable<Asn1Primitive> {
+sealed class Asn1String(
```
JesusMcCloud, line 40: Add the info to the API docs about defaulting to UTF-8 and the possibility to handle different charsets by working with `rawValue`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,193 @@
...
+    open val rawValue: ByteArray
```
JesusMcCloud: need not be open

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
JesusMcCloud: In general, it's fine if all the constructors throw here for validation errors, but they need to be annotated with `@Throws` and API docs with `@throws` that explain when and why exceptions are thrown

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,193 @@
...
         override val tag = BERTags.UTF8_STRING.toULong()
+        constructor(value: String) : this(value.encodeToByteArray())
+
+        init {
+            if (value.contains('\uFFFD')) throw Asn1Exception("Input contains invalid chars: '$value'")
+        }
     }
 
     /**
      * UNIVERSAL STRING (unchecked)
```
JesusMcCloud: fine, because it is complex and discouraged

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
JesusMcCloud: I am conflicted and need more input on this: I'd say, we want to add a flag to a (private/internal) default constructor that disables sanity checks, so we remain able to parse botched strings contained in ASN.1 structures (they are out there, I'm certain of it!), but prevent programmatically creating illegal strings. In order to communicate this to the user, I'd propose the following contract:

- add a flag to the root string class (`isValidated`) that indicates whether validation has taken place and clearly communicate its semantics, also explaining that ASN.1 decoded strings are never validated to be able to accept illegal encodings. this is the same flag that is specified in the private constructor i mentioned before
- introduce an abstract `isValid` property that lazily performs the validation logic and returns
    - `true` if validation was successful
    - `false` if not
    - `null` if no validation is possible/implemented

- We keep/add the `@throws` for the public constructors that always validate, because we do not want to let users create invalid strings. If they really insist on shooting themselves in the foot, they should manually create an ASN.1 primitive with an illegal string.

@iaik-jheher @nodh  any thoughts?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
```diff
@@ -328,25 +329,76 @@ inline fun Asn1Primitive.decodeToFloatOrNull(assertTag: Asn1Element.Tag = Asn1El
...
+    runRethrowing { decode(assertTag) { Asn1String.Unrestricted(content) } }
+
+@Throws(Asn1Exception::class)
+inline fun Asn1Primitive.decodeToVideotexString(assertTag: Asn1Element.Tag = Asn1Element.Tag.STRING_VIDEOTEX) =
+    runRethrowing { decode(assertTag) { Asn1String.Videotex(content) } }
+
+
 /**
  * Decodes this [Asn1Primitive]'s content into a String.
  * @throws [Asn1Exception] all sorts of exceptions on invalid input
```
JesusMcCloud, line 459: add a note to the api docs that the type of the string will be determined based on the tag and refer to the specific methods you introduces above that allow tag overrides. also: add the missing `@Throws` annotation

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
StjepanovicSrdjan: Done in https://github.com/a-sit-plus/signum/pull/345/commits/3ccfe7dd827d63dc1d9ffe8c11cb5a35d51e6d87

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/encoding/Asn1Decoding.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/345/commits/3ccfe7dd827d63dc1d9ffe8c11cb5a35d51e6d87 and  https://github.com/a-sit-plus/signum/pull/345/commits/1a98675000adaed08dbab1606b468254da7c50f8

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+ * ASN.1 String class used as wrapper do discriminate between different ASN.1 string types
+ * By default, the string value is decoded using UTF-8. If a different charset or custom decoding
+ * is needed, the [rawValue] property can be used directly.
+ *
+ * Constructor distinctions:
+ *
+ * 1. **Public constructors (String input)**
+ *    - These constructors validate the input against the allowed character set for that
+ *      specific ASN.1 string type (e.g., PrintableString, IA5String).
+ *    - This ensures that strings created by library users are always valid.
```
JesusMcCloud: > created programmatically

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+ *
+ * 2. **Internal/raw constructors (ByteArray input)**
+ *    - These constructors don't performing validation.
+ *    - Used internally for decoding ASN.1 strings to be able to accept illegal encodings.
```
JesusMcCloud: rephrase to be less focused on the implementations. Rather something along the lines of:
> To enable parsing of non-compliant strings (i.e. as part of certificate subjects) without exploding, every String is internally represented as `ByteArray` and not validated during decoding from ASN.1. The [isValid] property indicates whether the bytes contained in an ASN.1 String object type are valid according to the validation rules of that type.

I'd also rename the `isValidated` constructor flag to `performValidation` or something similar to prevent mixups when working on the code due to the similarity with `isValid`

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
  */
-sealed class Asn1String : Asn1Encodable<Asn1Primitive> {
+sealed class Asn1String(
+    val rawValue: ByteArray,
+    val isValidated: Boolean
+) : Asn1Encodable<Asn1Primitive> {
     abstract val tag: ULong
-    abstract val value: String
+    val value: String by lazy { String.decodeFromAsn1ContentBytes(rawValue) }
```
JesusMcCloud, line 51: Add API docs here indicating that this is always the UTF-8 interpretation of [rawValue] as per stdlib's [decodeToString]. Directly link to that function, then it should be crystal clear. The API docs of `String.decodeFromAsn1ContentBytes` would also benefit from being augmented with that information.

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        constructor(value: String) : this(value.encodeToByteArray(), true) {
+            if (!isValid) throw Asn1Exception("Input contains invalid chars: '$value'")
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
      * UNIVERSAL STRING (unchecked)
```
JesusMcCloud: explicitly mention that no validation is implemented (or even possible?) for universal strings.

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
      */
-    class Universal(override val value: String) : Asn1String() {
+    class Universal private constructor(
+        rawValue: ByteArray,
+        isValidated: Boolean
+    ) : Asn1String(rawValue, isValidated) {
         override val tag = BERTags.UNIVERSAL_STRING.toULong()
+
+        override val isValid: Boolean? = null
```
JesusMcCloud, line 100: apidocs should mention that this will always be null for universal string, as no validation logic is present/possible

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
JesusMcCloud: also don't mention constructors explicitly (see below comment).

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+     * - `true`: validation succeeded
+     * - `false`: validation failed
+     * - `null`: no validation implemented
+     */
+    abstract val isValid: Boolean?
+
 
     /**
      * UTF8 STRING (verbatim String)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        constructor(value: String) : this(value.encodeToByteArray(), false)
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * VISIBLE STRING (no checks)
+     * VISIBLE STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * IA5 STRING (no checks)
+     * IA5 STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * TELETEX STRING (no checks)
+     * TELETEX STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * BMP STRING (no checks)
+     * BMP STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * PRINTABLE STRING (checked)
+     * GENERAL STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+            if (!isValid) throw Asn1Exception("Input contains invalid chars: '$value'")
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
+    }
+
+    /**
+     * GRAPHIC STRING (checked)
      * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+        override val tag = BERTags.UNRESTRICTED_STRING.toULong()
+
+        override val isValid: Boolean? = null
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
+    }
+
+    /**
+     * VIDEOTEX STRING (no checks)
```
JesusMcCloud, line 276: also be more explicit about no implemented checks for this string type

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+     */
+    class Videotex private constructor(
+        rawValue: ByteArray,
+        isValidated: Boolean
+    ) : Asn1String(rawValue, isValidated) {
+        override val tag = BERTags.VIDEOTEX_STRING.toULong()
+
+        override val isValid: Boolean? = null
```
JesusMcCloud, line 288: same as with universal string: explicitly mention that no validation is performed

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
+    }
+
+    /**
+     * PRINTABLE STRING (checked)
+     * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,323 @@
...
+            if (!isValid) throw Asn1Exception("Input contains invalid chars: '$value'")
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
      * NUMERIC STRING (checked)
      * @throws Asn1Exception if illegal characters are provided
```
JesusMcCloud: this ctor does not throw, so this api doc line needs to be moved to the public ctor

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/345/commits/5cfe143a848f8540ab338e670b82cbaf30ea0054

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/345/commits/5cfe143a848f8540ab338e670b82cbaf30ea0054

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/345/commits/5cfe143a848f8540ab338e670b82cbaf30ea0054

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,349 @@
...
 /**
- * ASN.! String class used as wrapper do discriminate between different ASN.1 string types
+ * ASN.1 String class used as wrapper do discriminate between different ASN.1 string types
+ * By default, the string value is decoded using UTF-8. If a different charset or custom decoding
+ * is needed, the [rawValue] property can be used directly.
+ *
+ * To enable parsing of non-compliant strings without exploding, every String is internally represented
+ * as raw [ByteArray] and not validated during decoding from ASN.1.
+ * The [isValid] property indicates whether the bytes contained in an ASN.1 String object type are valid
+ * according to the validation rule of that type
```
JesusMcCloud:
```suggestion
 * according to the validation rules of that type.
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,349 @@
...
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * UNIVERSAL STRING (unchecked)
+     * UNIVERSAL STRING (no checks)
+     * Validation is not implemented. This string format is not recommended
+     * and is often replaced with UTF-8.
```
JesusMcCloud:
```suggestion
     * Validation is not implemented. This string format is deprecated for HTTPS certificates and its use in generally discouraged in favor of UTF-8 strings (see [Asn1String.UTF8]).
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,349 @@
...
+            if (!isValid) throw Asn1Exception("Input contains invalid chars: '$value'")
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * TELETEX STRING (no checks)
+     * TELETEX STRING (checked)
```
JesusMcCloud:
```suggestion
     * TELETEX STRING (checked).
     *  This string format is deprecated for HTTPS certificates and its use in generally discouraged in favor of UTF-8 strings (see [Asn1String.UTF8]).
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,349 @@
...
+        }
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
     }
 
     /**
-     * BMP STRING (no checks)
+     * BMP STRING (unchecked)
+     * Validation is not implemented. This string format is not recommended
```
JesusMcCloud:
```suggestion
     * BMP STRING (unchecked).
     * Validation is not implemented. This string format is deprecated for HTTPS certificates and its use in generally discouraged in favor of UTF-8 strings (see [Asn1String.UTF8]).
```

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
```diff
@@ -1,86 +1,349 @@
...
+
+        constructor(value: String) : this(value.encodeToByteArray(), false)
+
+        @PublishedApi
+        internal constructor(rawValue: ByteArray) : this(rawValue, false)
+    }
+
+    /**
+     * VIDEOTEX STRING (no checks)
+     * Validation is not implemented. This type is no longer used.
```
JesusMcCloud:
```suggestion
     * Validation is not implemented. This type is no longer used in practice.
```

indispensable-asn1/src/jvmTest/resources/asn1strings.json
JesusMcCloud, line 1: I think the test String JSON is fine for now, since we don't validate on parsing. Open an issue for more systematic, more complete string validation tests and reference this comment (and this file here). There's almost certainly something out there. maybe bouncy castle test?

indispensable-asn1/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1String.kt
JesusMcCloud: I'm waiting for @nodh to chime in on this before I resolve it

JesusMcCloud (approved):
LGTM pending

- my suggestions (that you can just apply directly through the Web UI, if you agree, since we'll be squashing anyhow)
- the creation of the issue for the string tests as per my comment
- @nodh's approval

indispensable-asn1/src/jvmTest/resources/asn1strings.json
StjepanovicSrdjan: I created the issue, I will dig deeper for some better tests.

Issue #329, opened 2025-07-22
Unify `entries`


JesusMcCloud:
We're using `entries` on various companion objects to emulate enum-style exhaustive listing. However, we do so inconsistently. At times it is a list, then it is a set, the return types are sometimes declared as `Iterable` other times its a set, or a list. In short: it's a mess.  
This concerns all indispensable modules

Entries should be:
* sets
* declared as sets
* lazily evaluated to prevent initialization cycles. Especially with Kotlin 2.2 this is becoming a very serious concern

PR #348, opened 2025-09-01
Enhancement/unify entries

StjepanovicSrdjan:
Unifying entries #329 
- lazy evaluated sets

### DIFF
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsAlgorithm.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsAlgorithm.kt
@@ -50,7 +50,7 @@ sealed class JwsAlgorithm(override val identifier: String) :
                 }
 
             companion object : Enumeration<Signature.EC> {
-                override val entries: List<EC> by lazy {
+                override val entries: Collection<EC> by lazy {
                     listOf(
                         ES256,
                         ES384,
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/DataIntegrityAlgorithm.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/DataIntegrityAlgorithm.kt
@@ -10,7 +10,7 @@ import at.asitplus.signum.Enumeration
  */
 sealed interface DataIntegrityAlgorithm : Enumerable {
     companion object : Enumeration<DataIntegrityAlgorithm> {
-        override val entries: Iterable<DataIntegrityAlgorithm> by lazy { MessageAuthenticationCode.entries.toSet() + SignatureAlgorithm.entries }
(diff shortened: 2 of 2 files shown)


JesusMcCloud (changes requested):
This looks good, but given this recurring pattern, we should probably introduce an interface for this, as it will be commonplace once we make everything extensible. I'm thinking twin interfaces similar to Asn1Encodable/Decodable, so we can easily write unit tests that ensure that companion objects are present and the types align. We have a basis for such tests already, but those tests are flaky and do not work well for class hierarchies. @iaik-jheher any thoughts?

StjepanovicSrdjan:
> This looks good, but given this recurring pattern, we should probably introduce an interface for this, as it will be commonplace once we make everything extensible. I'm thinking twin interfaces similar to Asn1Encodable/Decodable, so we can easily write unit tests that ensure that companion objects are present and the types align. We have a basis for such tests already, but those tests are flaky and do not work well for class hierarchies. @iaik-jheher any thoughts?

Would that be something like:

`interface Registrable<T>`

`interface Registry<T : Registrable<T>> {
    val entries: Set<T>
}`

and if so, how would we handle sealed class hierarchy where every subclass have entries (e.g. `JwsAlgorithm`)?

indispensable-asn1/build.gradle.kts
```diff
@@ -65,6 +65,7 @@ kotlin {
             )
 
             dependencies {
+                implementation(project(":internals"))
```
JesusMcCloud: I don't think having that in internals will work, since we want to publicly expose it. i think the indispensable module is the right place for that interface tandem, since we need it in all modules that depend on indispensable (wich are all others,  and we don't need it in indispensale-asn1, wich is the only one that does not depend on it)

JesusMcCloud (changes requested):
Looking good!

@iaik-jheher could you expand your reflection-powered testcase to cover this pattern before we merge this, wo we have automatic tests? IIRC the current tests for algorithms are brittle and can't be used as a template as-is.

indispensable-asn1/build.gradle.kts
StjepanovicSrdjan: Actually, we use the interface in Asn1Elements.kt from indispensable-asn1, so that was the only reason I didn't place it in indispensable module.

JesusMcCloud: I see. thing is: internals is for internal use and not exposed. I mean, if it still works from a consuming project (which i doubt), it can stay. otherwise, i think it must go into the asn.1 module.

StjepanovicSrdjan: Done in https://github.com/a-sit-plus/signum/pull/348/commits/586550c5dfb645a4bdbd77045cade964c3b94e6c

JesusMcCloud:
add API docs on both interfaces (a short two-liner, also mentioning that entries do not guarantee any order) and add a changelog entry specifically highlighting those classes that changed, e.g. from List/Collection/… to Set semantics. I'll adapt the test cases Jakob made to match this pattern in parallel

PR #353, opened 2025-09-11
Enhancement: GeneralName support

StjepanovicSrdjan:
- Refactored `AlternativeNames` for SAN/IAN extraction
- Removed detailed parsing of individual name types; now delegates decoding to `GeneralName` and subclasses

### DIFF
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesConstrainsTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesConstrainsTest.kt
@@ -6,7 +6,8 @@ import at.asitplus.signum.indispensable.asn1.KnownOIDs
 import at.asitplus.signum.indispensable.asn1.location
 import at.asitplus.signum.indispensable.pki.AttributeTypeAndValue
 import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName
-import at.asitplus.test.FreeSpec
+import at.asitplus.testballoon.invoke
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.matchers.shouldBe
 
 /**
@@ -14,7 +15,7 @@ import io.kotest.matchers.shouldBe
  * https://github.com/bcgit/bc-java/blob/126ac9e14a0f56fae088973a777f1f90a521fd82/prov/src/test/java/org/bouncycastle/jce/provider/test/PKIXNameConstraintsTest.java
  */
 @OptIn(ExperimentalPkiApi::class)
-class GeneralNamesConstrainsTest : FreeSpec ({
+val GeneralNamesConstrainsTest by testSuite{
 
     fun <T : GeneralNameOption> testGeneralNameConstraints(
         name: String,
@@ -188,7 +189,7 @@ class GeneralNamesConstrainsTest : FreeSpec ({
         narrowNames = narrowIps,
         diffTypeOther = dummyOtherRFC,
     )
-})
+}
 
 
 fun x500NameFromString(dn: String): X500Name {
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500NameParsingTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500NameParsingTest.kt
@@ -1,13 +1,14 @@
 package at.asitplus.signum.indispensable.pki.generalNames
 
-import at.asitplus.test.FreeSpec
+import at.asitplus.testballoon.invoke
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.matchers.shouldBe
 
 /**
  * Adapted from BouncyCastle's X500NameTest:
  * https://github.com/bcgit/bc-java/blob/main/core/src/test/java/org/bouncycastle/asn1/test/X500NameTest.java
  */
-class X500NameParsingTest : FreeSpec ({
+val X500NameParsingTest by testSuite {
 
     fun assertCanonical(input: String, expected: String) {
         val name = X500Name.fromString(input)
@@ -65,10 +66,14 @@ class X500NameParsingTest : FreeSpec ({
     }
 
     "testQuotedEqualityAndCase" {
-        assertCanonical("CN=\"  CA1 -   CP.04.03\", OU=Testing, O=U.S. Government, C=US",
-            "cn=ca1 - cp.04.03,ou=testing,o=u.s. government,c=us")
-        assertCanonical("CN=\"ca1 - CP.04.03  \", OU=Testing, O=U.S. Government, C=US",
-            "cn=ca1 - cp.04.03,ou=testing,o=u.s. government,c=us")
+        assertCanonical(
+            "CN=\"  CA1 -   CP.04.03\", OU=Testing, O=U.S. Government, C=US",
+            "cn=ca1 - cp.04.03,ou=testing,o=u.s. government,c=us"
+        )
+        assertCanonical(
+            "CN=\"ca1 - CP.04.03  \", OU=Testing, O=U.S. Government, C=US",
+            "cn=ca1 - cp.04.03,ou=testing,o=u.s. government,c=us"
+        )
     }
 
     "testCaseInsensitiveEquality" {
@@ -94,4 +99,4 @@ class X500NameParsingTest : FreeSpec ({
         assertCanonical("CN=#GG11", "cn=\\#gg11")
     }
 
-})
+}
--- a/indispensable/src/commonTest/kotlin/GeneralNameDecodingTest.kt
+++ b/indispensable/src/commonTest/kotlin/GeneralNameDecodingTest.kt
@@ -1,12 +1,13 @@
 package at.asitplus.signum
 import at.asitplus.signum.indispensable.pki.X509Certificate
 import at.asitplus.signum.indispensable.pki.generalNames.GeneralNameOption
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.assertions.throwables.shouldNotThrowAny
-import io.kotest.core.spec.style.FreeSpec
 import io.kotest.matchers.shouldBe
+import at.asitplus.testballoon.invoke
 
 @OptIn(ExperimentalPkiApi::class)
-open class GeneralNameDecodingTest : FreeSpec({
+val GeneralNameDecodingTest by testSuite{
 
     "Alternative Names Decoding" {
         val sanRFC822namesPem = "-----BEGIN CERTIFICATE-----\n" +
@@ -275,4 +276,4 @@ open class GeneralNameDecodingTest : FreeSpec({
         generalNames = cert.tbsCertificate.subjectAlternativeNames?.generalNames
         generalNames?.forEach { it.name.type shouldBe GeneralNameOption.NameType.DIRECTORY }
     }
-})
+}
--- a/indispensable/src/commonTest/kotlin/UtilTest.kt
+++ b/indispensable/src/commonTest/kotlin/UtilTest.kt
@@ -9,9 +9,6 @@ import com.ionspin.kotlin.bignum.integer.BigInteger
 import at.asitplus.testballoon.invoke
 import de.infix.testBalloon.framework.testSuite
 import io.kotest.matchers.shouldBe
-import de.infix.testBalloon.framework.TestConfig
-import kotlin.time.Duration.Companion.minutes
-import de.infix.testBalloon.framework.testScope
 
 val UtilTest by testSuite {
    
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateJvmTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateJvmTest.kt
@@ -162,6 +162,7 @@ val X509CertificateJvmTest by testSuite {
     }
 
     "Certificate can be parsed" {
+        val keyPair: KeyPair = keyGen.genKeyPair()
         val ecPublicKey = keyPair.public as ECPublicKey
         val keyX = ecPublicKey.w.affineX.toByteArray().ensureSize(ecCurve.coordinateLength.bytes)
         val keyY = ecPublicKey.w.affineY.toByteArray().ensureSize(ecCurve.coordinateLength.bytes)
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DefaultValidationTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DefaultValidationTest.kt
@@ -3,10 +3,11 @@ package at.asitplus.signum.indispensable.pki.generalNames
 import at.asitplus.signum.ExperimentalPkiApi
 import at.asitplus.signum.indispensable.asn1.Asn1Element
 import at.asitplus.signum.indispensable.asn1.encoding.parse
-import at.asitplus.test.FreeSpec
+import at.asitplus.testballoon.invoke
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.matchers.shouldBe
 
-class DefaultValidationTest : FreeSpec ({
+val DefaultValidationTest by testSuite{
     val bytes = byteArrayOf(-96, 30, 6, 3, 42, 3, 4, -96, 23, 12, 21, 115, 111, 109, 101, 32, 111, 116, 104, 101, 114, 32, 105, 100, 101, 110, 116, 105, 102, 105, 101, 114)
 
     "EDIPartyName without custom validation " {
@@ -38,4 +39,4 @@ class DefaultValidationTest : FreeSpec ({
         val validated = name.createValidatedCopy { true }
         validated.isValid shouldBe true
     }
-})
+}
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesEqualityTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesEqualityTest.kt
@@ -7,11 +7,12 @@ import at.asitplus.signum.indispensable.asn1.Asn1Element
 import at.asitplus.signum.indispensable.asn1.Asn1String
 import at.asitplus.signum.indispensable.asn1.encoding.parse
 import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName
-import at.asitplus.test.FreeSpec
+import at.asitplus.testballoon.invoke
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
 
-class GeneralNamesEqualityTest : FreeSpec ({
+val GeneralNamesEqualityTest by testSuite{
 
     val bytes = byteArrayOf(-96, 30, 6, 3, 42, 3, 4, -96, 23, 12, 21, 115, 111, 109, 101, 32, 111, 116, 104, 101, 114, 32, 105, 100, 101, 110, 116, 105, 102, 105, 101, 114)
     val changedBytes = byteArrayOf(-95, 31, 6, 3, 42, 3, 4, -96, 24, 12, 22, 115, 111, 109, 101, 32, 101, 100, 105, 116, 101, 100, 32, 105, 100, 101, 110, 116, 105, 102, 105, 101, 114)
@@ -236,4 +237,4 @@ class GeneralNamesEqualityTest : FreeSpec ({
         x500multi1 shouldNotBe x500multi3
         x500name1 shouldNotBe x500multi1
     }
-})
+}
--- a/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/RelativeDistinguishedNameParsingTest.kt
+++ b/indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/RelativeDistinguishedNameParsingTest.kt
@@ -4,12 +4,13 @@ import at.asitplus.signum.indispensable.asn1.Asn1String
 import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName
 import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName.Companion.splitFirstUnescaped
 import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName.Companion.splitRespectingEscapeAndQuotes
-import at.asitplus.test.FreeSpec
+import at.asitplus.testballoon.invoke
+import de.infix.testBalloon.framework.testSuite
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.matchers.comparables.shouldBeGreaterThan
 import io.kotest.matchers.shouldBe
 
-class RelativeDistinguishedNameParsingTest : FreeSpec({
+val RelativeDistinguishedNameParsingTest  by testSuite{
 
     "splitRespectingEscapeAndQuotes should split simple RDN" {
         val input = "CN=John+O=Company"
@@ -115,4 +116,4 @@ class RelativeDistinguishedNameParsingTest : FreeSpec({
             recombined shouldBe dn
         }
     }
-})
+}


indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500Name.kt
```diff
@@ -0,0 +1,91 @@
...
+import at.asitplus.signum.indispensable.asn1.Asn1Primitive
+import at.asitplus.signum.indispensable.asn1.Asn1Sequence
+import at.asitplus.signum.indispensable.asn1.decodeRethrowing
+import at.asitplus.signum.indispensable.asn1.encoding.Asn1
+import at.asitplus.signum.indispensable.asn1.encoding.asAsn1String
+import at.asitplus.signum.indispensable.pki.AttributeTypeAndValue
+import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName
+
+data class X500Name(
+    val relativeDistinguishedNames: List<RelativeDistinguishedName>,
```
JesusMcCloud, line 13: add a secondary constructor that takes a single value instead of a list for convenience

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -34,56 +48,51 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
 }
 
 //TODO: value should be Asn1Primitive???
-sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
-    abstract val value: Asn1Element
+open class AttributeTypeAndValue(
+    override val oid: ObjectIdentifier,
+    val value: Asn1Element
```
JesusMcCloud: please check if this is guaranteed to be a primitive according to X.509. either way, the TODO needs to go

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
```diff
@@ -0,0 +1,134 @@
...
+import at.asitplus.signum.indispensable.asn1.TagClass
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToIa5String
+import at.asitplus.signum.indispensable.asn1.runRethrowing
+import kotlinx.io.IOException
+
+data class DNSName internal constructor(
+    val value: Asn1String.IA5,
+    val allowWildcard: Boolean = true,
+    override val type: GeneralNameOption.NameType = GeneralNameOption.NameType.DNS,
+) : GeneralNameOption, Asn1Encodable<Asn1Primitive> {
```
JesusMcCloud, line 18: I think the generalnames should also get an `isValidated` (and most many of them will delegate to the underlying Asn1String's `isValidated`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
```diff
@@ -0,0 +1,134 @@
...
+            return runRethrowing {
+                DNSName(
+                    type = GeneralNameOption.NameType.DNS,
+                    value = src.decodeToIa5String(tag)
+                )
+            }
+        }
+
+        fun fromString(value: String, allowWildcard: Boolean = true): DNSName {
+            val alphaDigits = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
```
JesusMcCloud: so this is even more constrained than IA5 string?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
```diff
@@ -0,0 +1,42 @@
...
+    companion object : Asn1Decodable<Asn1Element, OtherName> {
+        override fun doDecode(src: Asn1Element): OtherName {
+            if (src !is Asn1ExplicitlyTagged) throw Asn1StructuralException("Invalid ediPartyName Alternative Name found: ${src.toDerHexString()}")
+
+            src.also { it ->
+                if (it.children.size > 2) throw Asn1StructuralException("Invalid partyName Alternative Name found (>2 children): ${it.toDerHexString()}")
+                if (it.children.find { it.tag.tagValue != GeneralNameOption.NameType.OTHER.value && it.tag.tagValue != GeneralNameOption.NameType.RFC822.value } != null) throw Asn1StructuralException(
+                    "Invalid partyName Alternative Name found (illegal implicit tag): ${it.toDerHexString()}"
+                )
+//                //TODO: strict string parsing
```
JesusMcCloud: I think with the `isValidated` pattern, this TODO can move into the body of wherever a placeholder validation methods will live

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralName.kt
```diff
@@ -0,0 +1,79 @@
...
+                NameType.entries.firstOrNull { it.value == tagValue }
+        }
+    }
+
+    enum class ConstraintResult {
+        DIFF_TYPE,     // Different type, no constraint
+        MATCH,         // Exact match
+        NARROWS,       // Input narrows this name
+        WIDENS,        // Input widens this name
+        SAME_TYPE;      // Same type, but no match/narrow/widen
```
JesusMcCloud: nitpick: one space too many

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+    val address: IpAddress<*>,
+    val networkV4: IpNetwork.V4? = null,
+    val networkV6: IpNetwork.V6? = null,
+    override val type: GeneralNameOption.NameType = GeneralNameOption.NameType.IP
+) : GeneralNameOption, Asn1Encodable<Asn1Primitive> {
+
+    override fun encodeToTlv(): Asn1Primitive {
+        val bytes = when {
+            networkV4 != null -> networkV4.address.octets + networkV4.netmask
+            networkV6 != null -> networkV6.address.octets + networkV6.netmask
```
JesusMcCloud: is this an X.509-specific encoding or is this defined in https://www.rfc-editor.org/rfc/rfc4632 or one of its descendents? Because if it is, this should go into CIDRE

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+        private fun createNetworkV4(bytes: ByteArray): IPAddressName {
+            val address = IpAddress.V4(bytes.copyOfRange(0, 4))
+            val prefix = bytes.copyOfRange(4, 8).toPrefix()
+            return IPAddressName(address, networkV4 = IpNetwork.V4(address, prefix))
+        }
+
+        private fun createNetworkV6(bytes: ByteArray): IPAddressName {
+            val address = IpAddress.V6(bytes.copyOfRange(0, 16))
+            val prefix = bytes.copyOfRange(16, 32).toPrefix()
+            return IPAddressName(address, networkV6 = IpNetwork.V6(address, prefix))
```
JesusMcCloud: Same here: if this is standard, (or otherwise established as de-facto standard), this should go into CIDRE.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+import at.asitplus.signum.indispensable.asn1.Asn1Decodable
+import at.asitplus.signum.indispensable.asn1.Asn1Encodable
+import at.asitplus.signum.indispensable.asn1.Asn1Primitive
+import at.asitplus.signum.indispensable.asn1.encoding.encodeToAsn1OctetStringPrimitive
+import kotlinx.io.IOException
+
+data class IPAddressName(
+    val address: IpAddress<*>,
+    val networkV4: IpNetwork.V4? = null,
+    val networkV6: IpNetwork.V6? = null,
```
JesusMcCloud: should be a single `IpNetwork<*,*>`, shouldn't it?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+        result = 31 * result + (networkV6?.hashCode() ?: 0)
+        result = 31 * result + type.hashCode()
+        return result
+    }
+
+    override fun constrains(input: GeneralNameOption?): GeneralNameOption.ConstraintResult {
+        if (input !is IPAddressName) return GeneralNameOption.ConstraintResult.DIFF_TYPE
+        if (this == input) return GeneralNameOption.ConstraintResult.MATCH
+
+        if ((address is IpAddress.V4 && input.address is IpAddress.V4 && networkV4 == null && input.networkV4 == null) ||
```
JesusMcCloud: you can gretly simplify this by just checking for `.family` or using `.isSameFamily()`. also having only a single network

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+        if (networkV6 != null && input.networkV6 != null) {
+            val thisNet = networkV6
+            val otherNet = input.networkV6
+            return when {
+                thisNet == otherNet -> GeneralNameOption.ConstraintResult.MATCH
+                thisNet.contains(otherNet) -> GeneralNameOption.ConstraintResult.WIDENS
+                otherNet.contains(thisNet) -> GeneralNameOption.ConstraintResult.NARROWS
+                else -> GeneralNameOption.ConstraintResult.SAME_TYPE
+            }
```
JesusMcCloud: can also be removed, it its a single network in the class

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,145 @@
...
+        if (input.networkV4 != null && address is IpAddress.V4) {
+            return if (input.networkV4.contains(address))
+                GeneralNameOption.ConstraintResult.WIDENS
+            else GeneralNameOption.ConstraintResult.SAME_TYPE
+        }
+        if (input.networkV6 != null && address is IpAddress.V6) {
+            return if (input.networkV6.contains(address))
+                GeneralNameOption.ConstraintResult.WIDENS
+            else GeneralNameOption.ConstraintResult.SAME_TYPE
+        }
```
JesusMcCloud: same here

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/UriName.kt
```diff
@@ -0,0 +1,143 @@
...
+import at.asitplus.signum.indispensable.asn1.Asn1Element
+import at.asitplus.signum.indispensable.asn1.Asn1Encodable
+import at.asitplus.signum.indispensable.asn1.Asn1Primitive
+import at.asitplus.signum.indispensable.asn1.Asn1String
+import at.asitplus.signum.indispensable.asn1.TagClass
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToIa5String
+import at.asitplus.signum.indispensable.asn1.runRethrowing
+import kotlinx.io.IOException
+
+data class UriName(
```
JesusMcCloud: check if we want https://klibs.io/project/eygraber/uri-kmp. could be overkill, but i'd assume it is well-tested and has a solid parser

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralName.kt
```diff
@@ -0,0 +1,79 @@
...
+
+import at.asitplus.signum.indispensable.asn1.Asn1Decodable
+import at.asitplus.signum.indispensable.asn1.Asn1Element
+import at.asitplus.signum.indispensable.asn1.Asn1Encodable
+import at.asitplus.signum.indispensable.asn1.Asn1Exception
+
+interface GeneralNameOption {
+
+    enum class NameType(val value: ULong) {
+        OTHER(0u), RFC822(1u), DNS(2u), X400(3u), DIRECTORY(4u), EDI(5u), URI(6u), IP(7u), OID(8u);
```
JesusMcCloud, line 20: well if these are all the ones that exist, we can make it a sealed interface

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: According to [RFC5280](https://www.rfc-editor.org/rfc/rfc5280#section-7.1) value can be any type.
`AttributeValue ::= ANY -- DEFINED BY AttributeType`
But I think in practice Asn1Primitives are always used for DNs. I don't think switching to primitive will cause any errors, but if we want to  stick strictly to the RFC, we can leave it as `Asn1Element`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
StjepanovicSrdjan: Yes, `DNSName` validate on higher semantic level. E.g. empty strings, leading or trailing dots, empty labels, wildcard etc.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: then the TODO comment needs to go and everything needs to stay as it it

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
JesusMcCloud: i like it being part of CIDRE

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/UriName.kt
StjepanovicSrdjan: used in https://github.com/a-sit-plus/signum/pull/353/commits/7e9673ead4c3e035d31b038d865bfe74bfa1617b

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: done in https://github.com/a-sit-plus/signum/pull/353/commits/b9b7373fcb9ad3db0c9e1ab605e9d578f1a7e287

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353#discussion_r2375263236

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353#discussion_r2375263236

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353#discussion_r2375263236

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: waiting for https://github.com/a-sit-plus/cidre/pull/4

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
StjepanovicSrdjan: done in https://github.com/a-sit-plus/signum/pull/353/commits/fb795caeb85bbf97e1fd63f99474e87294b2192d

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
StjepanovicSrdjan: done in https://github.com/a-sit-plus/signum/pull/353/commits/fb795caeb85bbf97e1fd63f99474e87294b2192d

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -120,6 +126,7 @@ sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
                 return@decodeRethrowing when (oid) {
                     CommonName.OID -> str.fold(onSuccess = { CommonName(it) }, onFailure = { CommonName(asn1String) })
                     Country.OID -> str.fold(onSuccess = { Country(it) }, onFailure = { Country(asn1String) })
+                    EmailAddress.OID -> str.fold(onSuccess = { EmailAddress(it) }, onFailure = { EmailAddress(asn1String) })
```
JesusMcCloud: don't all of these enforce validation, since their default constructors are called?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -128,11 +135,64 @@ sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
...
+                }
+        } ?: ("#" + (value.toDerHexString()))
+
+        return "$type=$valStr".lowercase()
+    }
+
+    /**
+     * Apply RFC 2253 escaping + whitespace rules.
+     */
+    private fun canonicalizeString(input: String): String {
```
JesusMcCloud: I assume we have test cases for both directions

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
```diff
@@ -0,0 +1,127 @@
...
+import at.asitplus.signum.indispensable.asn1.Asn1String
+import at.asitplus.signum.indispensable.asn1.TagClass
+import at.asitplus.signum.indispensable.asn1.encoding.decodeToIa5String
+import at.asitplus.signum.indispensable.asn1.runRethrowing
+import kotlinx.io.IOException
+
+data class DNSName internal constructor(
+    val value: Asn1String.IA5,
+    val allowWildcard: Boolean = true,
+    override val performValidation: Boolean = false,
```
JesusMcCloud: as discussed offline: should not be a property. neither here nor in the base interface. I won't be commenting on this throughout, because the compiler will complain, when its removed in the base interface anyhow

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralName.kt
```diff
@@ -0,0 +1,88 @@
...
+sealed interface GeneralNameOption {
+
+    /**
+     * Returns whether this name is valid:
+     * - `true`: validation succeeded
+     * - `false`: validation failed
+     * - `null`: no validation implemented
+     */
+    val isValid: Boolean?
+    val performValidation: Boolean
```
JesusMcCloud: should go

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
JesusMcCloud: this needs an init block that does validation depending on whether performValidation is true or not. then the secondary constructor is just a call to the default constructor, which invoked the validate function depending. you may then even set isValidated in the default constructor. maybe this also concerns Asn1String. but we can discuss offline.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/DNSName.kt
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+
+    override fun toString(): String {
+        return value.value
+    }
+
+    override fun constrains(input: GeneralNameOption?): GeneralNameOption.ConstraintResult {
+        if (input !is DNSName) {
```
JesusMcCloud: another general remark: the constraint validation must always fail for any object that is not validated (so `isValid` false or null should always cause a constraint violation being reported for if even only a single non-validated GeneralName is present). this is transitive: If a non-validated string is present -> constraint violation (even though `isValid==null` on an ASN.1 String means: we have not implemented checks).

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Always `true`, since creation of [IPAddressName] is only possible if the
+     * underlying [IpAddress] or [IpAddressAndPrefix] are valid
+     */
+    override val isValid: Boolean = true
+    /**
+     * Always `true`, since validation is done by [IpAddress] and [IpAddressAndPrefix]
+     */
+    override val performValidation: Boolean = true
```
JesusMcCloud: to make optional validation work as leniently as we want it, we need to store the raw bytes. also: double check it the nullable `network` is then semantically untuitive. Can a contract then smart-cast it to non-null, if `isValid` is true?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    companion object : Asn1Decodable<Asn1Primitive, IPAddressName> {
+
+        override fun doDecode(src: Asn1Primitive): IPAddressName {
+            val content = src.content
+            return when (content.size) {
+                4 -> IPAddressName(IpAddress.V4(content))
+                16 -> IPAddressName(IpAddress.V6(content))
+                8 -> createNetworkV4(content)
+                32 -> createNetworkV6(content)
```
JesusMcCloud: should use onstants from IpFamily

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/OtherName.kt
```diff
@@ -0,0 +1,42 @@
...
+data class OtherName (
+    val value: Asn1ExplicitlyTagged,
+    override val performValidation: Boolean = false,
+    override val type: GeneralNameOption.NameType = GeneralNameOption.NameType.OTHER
+): GeneralNameOption, Asn1Encodable<Asn1Element> {
+
+    /**
+     * Always `null`, since no validation logic is implemented
+     */
+    override val isValid: Boolean? = null
```
JesusMcCloud: will this work out the way I sketched it before? as in: false AND null will cause a constraint violation?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/UriName.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    @Throws(Asn1Exception::class)
+    constructor(value: String, allowWildcard: Boolean = false) : this(Asn1String.IA5(value), allowWildcard, true)
+
+    init {
+        if (host.value.isEmpty()) {
+            throw IOException("URI name cannot be empty")
+        }
+
+        val uri = Uri.parse(host.value)
```
JesusMcCloud: same as with ip address: if invalid, we need raw value

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X400AddressName.kt
```diff
@@ -0,0 +1,40 @@
...
+data class X400AddressName(
+    val value: Asn1Element,
+    override val performValidation: Boolean = false,
+    override val type: GeneralNameOption.NameType = GeneralNameOption.NameType.X400,
+) : GeneralNameOption, Asn1Encodable<Asn1Element> {
+
+    /**
+     * Always `null`, since no validation logic is implemented
+     */
+    override val isValid: Boolean? = null
```
JesusMcCloud: Ok we probably need to rethink null always returning a constraint violation. I need to think about this.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: You're right, fixed this legacy code in https://github.com/a-sit-plus/signum/pull/353/commits/bd53929b326f22bee7b47f20a8cfc93d145b74c2

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353/commits/d8092ceb131fa4e37c98251d28e012f9930f96bb, https://github.com/a-sit-plus/signum/pull/353/commits/a245064f5800e82e717a5d29dea185c0016214de and https://github.com/a-sit-plus/signum/pull/353/commits/043fcd91a1e49e48c284f7a2963295fc3592a3e7

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: No it can't, because `IpAddressName` can be valid with just the IpAddress without prefix, which means network can be null.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/UriName.kt
StjepanovicSrdjan: Actually property `host` is Asn1String.IA5 which is safely decoded without validation, so we don't need extra raw value.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353/commits/09eecc5dd8ee51a3b0ffc96596b9fe27512cfaa6

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,196 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
                 }
             }.let(::RelativeDistinguishedName)
         }
+
+        // Predefined RFC2253 keyword order
+        private val rfc2253Order = listOf(
+            "CN", "C", "L", "S", "ST", "O", "OU", "T", "IP", "STREET",
+            "DC", "DNQUALIFIER", "DNQ", "SURNAME", "GIVENNAME",
+            "INITIALS", "GENERATION", "EMAIL", "EMAILADDRESS",
+            "UID", "SERIALNUMBER"
```
JesusMcCloud: These should be grouped as constants somewhere

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,196 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+                        sb.append('\\').append(c) // preserve the escape
+                        escaped = false
+                    }
+                    c == '\\' -> escaped = true
+                    c == delimiter -> return listOf(sb.toString(), input.substring(i + 1))
+                    else -> sb.append(c)
+                }
+            }
+            return listOf(sb.toString())
+        }
```
JesusMcCloud: Can't we rely on some pre-existing functionality for that? I know serialization has it (think: escaping/unescaping string is virtually any test-based format) but I would expect this to be buried somewhere inside kotlin's stdlib, or any of the dependencies we already use here. When in doublt use Android studio, as it gives you free Tokens for Gemini-powered AI-assistance

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,196 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+                        sb.clear()
+                    }
+                    else -> sb.append(c)
+                }
+            }
+
+            parts.add(sb.toString().trim())
+            return parts
+        }
+
```
JesusMcCloud: same here

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,196 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
 }
 
-//TODO: value should be Asn1Primitive???
-sealed class AttributeTypeAndValue : Asn1Encodable<Asn1Sequence>, Identifiable {
-    abstract val value: Asn1Element
+open class AttributeTypeAndValue(
+    override val oid: ObjectIdentifier,
+    val value: Asn1Element,
+    val attrType: String = oid.toString(),
+    val performValidation: Boolean = false
```
JesusMcCloud: should not be `val`, should it?

JesusMcCloud: Also: I don't see it used anywhere

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
```diff
@@ -0,0 +1,42 @@
...
+            }
+            return EDIPartyName(src)
+        }
+    }
+
+    override fun toString(): String {
+        return value.prettyPrint()
+    }
+
+    override fun validatedCopy(checkIsValid: (GeneralNameOption) -> Boolean): EDIPartyName {
```
JesusMcCloud: - just call the lambda `validate`
- rename to `createValidatedCopy`
- make it possible to pass the validation lambda into the co8nstructor. I know the parser will never call it, but if the user wants to create this type and ensure validation, it should be possible without copying. Then this function would just delegate to that constructor.

Same goes for other GeneralNames that must rely on custom validation

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralName.kt
```diff
@@ -0,0 +1,123 @@
...
+    val type: NameType
+
+    fun constrains(input: GeneralNameOption?): ConstraintResult {
+        when {
+            input == null || this::class != input::class -> return ConstraintResult.DIFF_TYPE
+
+            isValid == null || input.isValid == null ->
+                throw IllegalArgumentException(
+                    "Validation for ${this::class.simpleName} has not been performed. " +
+                            "Use ${this::class.simpleName}.validatedCopy { /* validation lambda */ } to set isValid before calling constrains."
```
JesusMcCloud: rephrase a bit, to be more explicit about the fact that the class does not support validation OOTB and requires passing custom validation logic

JesusMcCloud (changes requested):

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: I did refactor the code to use regex, but still didn't find something pre-existing. It is now much cleaner, but not sure if it's satisfactory yet. https://github.com/a-sit-plus/signum/pull/353/commits/ae61ba7227217d30c33d9f7a41f3fd3a9a7c309b

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353/commits/13d3e8fe43f967208e06d64091eecda3bf23d6f8

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralName.kt
StjepanovicSrdjan: https://github.com/a-sit-plus/signum/pull/353/commits/13d3e8fe43f967208e06d64091eecda3bf23d6f8

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -18,6 +30,16 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+    val sortedAttrsAndValues by lazy {
+        if (attrsAndValues.size > 1) {
+            attrsAndValues.sortedWith(compareBy { atv ->
+                Rfc2253Constants.ORDER[atv.attrType.uppercase()] ?: Int.MAX_VALUE
+            })
+        } else {
+            attrsAndValues
+        }
+    }
+
```
JesusMcCloud: It has to be encoded in sorted order, to be canonicalised, right?!
Or is it legal to encode it in any order? Because if not, then the default constructor must be private, and keep everything as-is to properly re-encode stuff without resorting and the only accessible constructor must call the default ctor but sort/canonicalise everything, so we cannot construct some non-compliant data.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+) : Asn1Encodable<Asn1Sequence>, Identifiable {
+
+    /**
+     * Returns whether this string is valid:
+     * - `true`: validation succeeded
+     * - `false`: validation failed
+     * - `null`: no validation implemented
+     */
+    val isValid: Boolean? by lazy {
+        runCatching { Asn1String.decodeFromTlv(value.asPrimitive()).isValid }.getOrNull()
```
JesusMcCloud: catchingUnwrapped

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+    class EmailAddress internal constructor(
+        value: Asn1Element,
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
+
+        /**
+         * @throws Asn1Exception if illegal EmailAddress is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (!isValid!!) throw Asn1Exception("Invalid EmailAddress!")
```
JesusMcCloud: will NPE for non-validating string.
Also; another Ctor taking a plain kotlin string, validate it as UTF8-String and then call this ctor

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesConstrainsTest.kt
```diff
@@ -0,0 +1,204 @@
...
+
+import at.asitplus.signum.indispensable.asn1.Asn1String
+import at.asitplus.signum.indispensable.asn1.KnownOIDs
+import at.asitplus.signum.indispensable.asn1.location
+import at.asitplus.signum.indispensable.pki.AttributeTypeAndValue
+import at.asitplus.signum.indispensable.pki.RelativeDistinguishedName
+import at.asitplus.test.FreeSpec
+import io.kotest.matchers.shouldBe
+
+class GeneralNamesConstrainsTest : FreeSpec ({
```
JesusMcCloud: Can we test all those against JCA / BouncyCastle in addition to your tests?

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500NameParsingTest.kt
```diff
@@ -0,0 +1,93 @@
+package at.asitplus.signum.indispensable.pki.generalNames
+
+import at.asitplus.test.FreeSpec
+import io.kotest.matchers.shouldBe
+
+class X500NameParsingTest : FreeSpec ({
```
JesusMcCloud: can we also test those against bouncy castle / JCA?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+        value: Asn1Element,
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        /**
+         * @throws Asn1Exception if illegal CommonName is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (!isValid!!) throw Asn1Exception("Invalid CommonName!")
```
JesusMcCloud: will NPE for non-validating string.
Also; another Ctor taking a plain kotlin string, validate it as UTF8-String and then call this ctor

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+        value: Asn1Element,
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        /**
+         * @throws Asn1Exception if illegal Country is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (!isValid!!) throw Asn1Exception("Invalid Country!")
```
JesusMcCloud: will NPE for non-validating string.
Also; another Ctor taking a plain kotlin string, validate it as UTF8-String and then call this ctor

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+        value: Asn1Element,
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        /**
+         * @throws Asn1Exception if illegal Organization is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (!isValid!!) throw Asn1Exception("Invalid Organization!")
```
JesusMcCloud: will NPE for non-validating string.
Also; another Ctor taking a plain kotlin string, validate it as UTF8-String and then call this ctor

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -27,63 +49,158 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+        value: Asn1Element,
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        /**
+         * @throws Asn1Exception if illegal OrganizationalUnit is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (!isValid!!) throw Asn1Exception("Invalid OrganizationalUnit!")
```
JesusMcCloud: will NPE for non-validating string.
Also; another Ctor taking a plain kotlin string, validate it as UTF8-String and then call this ctor

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
```diff
@@ -0,0 +1,47 @@
...
+    override val type: GeneralNameOption.NameType = GeneralNameOption.NameType.OTHER
+): GeneralNameOption, Asn1Encodable<Asn1Element> {
+
+    constructor(value: Asn1ExplicitlyTagged) : this(value, null)
+
+    /**
+     * Creates a new [EDIPartyName] instance with validation applied at construction time.
+     * This constructor allows supplying a custom [validate] lambda that determines the value of [isValid].
+     */
+    constructor(value: Asn1ExplicitlyTagged, validate: (GeneralNameOption) -> Boolean) :
```
JesusMcCloud, line 22: Implement a test that:

1. errors out, because no custom validation logic is passed
2. then constructs one using `createValidatedCopy` with an always-true validation lambda and succeeds to show this works

Also for any other generalNameOptions without default validation logic, in case I missed them.

Do we need this for those that take arbitrary Asn1Strings too? (is that even legal)?

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesConstrainsTest.kt
StjepanovicSrdjan: Tests are based on the BouncyCastle tests

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500NameParsingTest.kt
StjepanovicSrdjan: Tests are based on the BouncyCastle tests

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
StjepanovicSrdjan: All GeneralNames that take Asn1String use IA5, so the validation is properly delegated to the Asn1String.IA5

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
StjepanovicSrdjan: I address this in https://github.com/a-sit-plus/signum/pull/353/commits/57d03b371f6360429546600739eb3a9a43996405. Now we prevent user from creating RDN with invalid order, but if RDN is parsed invalid we keep it as-is.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,14 +1,49 @@
...
+import at.asitplus.signum.indispensable.asn1.readOid
+import at.asitplus.signum.indispensable.asn1.runRethrowing
 
 /**
  * X.500 Name (used in X.509 Certificates)
  */
-data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndValue>) : Asn1Encodable<Asn1Set> {
+class RelativeDistinguishedName private constructor(
+    val attrsAndValues: List<AttributeTypeAndValue>,
+    performValidation: Boolean = false,
```
JesusMcCloud: I'd omit the default value, to prevent even the possibility of accidentally calling the wrong c'tor

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -20,70 +55,198 @@ data class RelativeDistinguishedName(val attrsAndValues: List<AttributeTypeAndVa
...
+    ) : AttributeTypeAndValue(OID, value, TYPE) {
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        /**
+         * @throws Asn1Exception if illegal Organization is provided
+         */
+        @Throws(Asn1Exception::class)
+        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray())) {
+            if (isValid == false) throw Asn1Exception("Invalid Organization!")
+        }
```
JesusMcCloud, line 208: seeing this pattern everywhere: can't we pull that up as a protected c'tor in the base class?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: what about this? https://github.com/Kotlin/kotlinx.serialization/blob/71b5d5ee99f953b60756c4aaa2ac8e12c3fae605/formats/json/commonMain/src/kotlinx/serialization/json/internal/JsonStreams.kt#L19

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/GeneralNamesConstrainsTest.kt
JesusMcCloud: then please add a comment to the source of the test

indispensable/src/jvmTest/kotlin/at/asitplus/signum/indispensable/pki/generalNames/X500NameParsingTest.kt
JesusMcCloud: same, please link to the original BC tests in a comment

JesusMcCloud (changes requested):
nitpicks and one open point for an offline elaboration

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/EDIPartyName.kt
JesusMcCloud: i have no Idea what I was talking about

PR #354, opened 2025-09-12
fix serialization of symmetric CoseKey

StjepanovicSrdjan:


indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/CoseKey.kt
```diff
@@ -484,6 +484,15 @@ object CoseKeySerializer : KSerializer<CoseKey> {
             if (k == null) throw IllegalArgumentException("Parameter k not optional for symmetric keys")
         }
 
+        constructor(src: CoseKey) : this(
+            src.type,
```
nodh: nitpick: I would add parameter names to increase readability and catch mix-ups

PR #366, opened 2025-10-22
Java 21

nodh:


.github/workflows/test-jvm-21.yml
```diff
@@ -0,0 +1,50 @@
...
+
+  # run the workflow once more **after** the PR has been merged
+  push:
+    branches:
+      - main
+      - development
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
JesusMcCloud, line 14: not running apple targets. intentional?

nodh: Just like the other JVM job

JesusMcCloud: silly me, you just want JVM tests

PR #374, opened 2025-10-30
Enhancement - dedicated x509 extensions

StjepanovicSrdjan:
* Introduced dedicated X509 extension classes
* `X509CertificateExtension` is now a base class
* Enables polymorphic decoding/encoding of extension types
    * `BasicConstraintsExtension`
    * `CertificatePoliciesExtension`
    * `InhibitAnyPolicyExtension`
    * `KeyUsageExtension`
    * `NameConstraintsExtension` and `GeneralSubtree`
    * `PolicyConstraintsExtension`
    * `PolicyMappingsExtension`

### DIFF
--- a/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/GeneralSubtree.kt
+++ b/indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/GeneralSubtree.kt
@@ -77,7 +77,7 @@ data class GeneralSubtrees(
 
     /**
      * Removes all redundant entries
-     * */
+     */
     @OptIn(ExperimentalPkiApi::class)
     private fun minimize(): GeneralSubtrees {
         val mutableTrees = trees.toMutableList()
@@ -127,14 +127,14 @@ data class GeneralSubtrees(
     }
 
     @ExperimentalPkiApi
-    fun unionWith(other: GeneralSubtrees?) {
-        other?.trees?.let { trees.addAll(it) }
+    fun unionWith(other: GeneralSubtrees) {
+        trees.addAll(other.trees)
         minimize()
     }
 
     /**
-     * Creates Subtree containing widest name of that type]
-     * */
+     * Creates Subtree containing widest name of that type
+     */
     private fun createWidestSubtree(name: GeneralNameOption): GeneralSubtree {
         return try {
             val newName = when (name.type) {
@@ -155,7 +155,7 @@ data class GeneralSubtrees(
 
     /**
      * Merges permitted NameConstraints
-     * */
+     */
     @ExperimentalPkiApi
     fun intersectAndReturnExclusions(other: GeneralSubtrees): GeneralSubtrees? {
         require(other.trees != null) { "other GeneralSubtrees must not be null" }


indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/GeneralSubtree.kt
```diff
@@ -0,0 +1,281 @@
...
+            }.toMutableList()
+            return GeneralSubtrees(trees)
+        }
+    }
+
+    /**
+     * Removes all redundant entries
+     * */
+    @OptIn(ExperimentalPkiApi::class)
+    private fun minimize(): GeneralSubtrees {
```
StjepanovicSrdjan, line 82: Didn't write dedicated tests for this function, since it is private and tested through public `unionWith()` and `intersectAndReturnExclusions()` methods. If necessary I can make it internal and write tests.

gradle/libs.versions.toml
```diff
@@ -12,6 +12,8 @@ testballoon = "0.6.3-K2.2.20-SNAPSHOT"
 asp = "20251023"
 cidre = "0.3.0"
 urikmp = "0.0.20"
+kotlinxCoroutinesCore = "1.10.2"
```
JesusMcCloud: no need. Conventions does that: https://github.com/a-sit-plus/gradle-conventions-plugin?tab=readme-ov-file#dependency-shorthands

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
```diff
@@ -38,19 +59,46 @@ data class X509CertificateExtension @Throws(Asn1Exception::class) private constr
 
     companion object : Asn1Decodable<Asn1Sequence, X509CertificateExtension> {
 
+        private val _registeredExtensionDecoders = MutableStateFlow(
```
JesusMcCloud, line 86: I think this again needs the generic pattern of keeping raw, structurally valid data and failing softly, if the contents are invalid. This is all very brainstormy, but maybe even using a dedicated `InvalidExtension` makes sense here:

- a reserved OID (e.g. BasicConstraints) is used, but the contents are bogus
- Parsing fails
- An more then zero `InvalidExtension`s are present in the resulting certificate
- We know that something is not koscher
- We can check for the presence of an OID in the list of `InvalidExtension`s that our cert path validator requires for valid certificates
- If so -> it will fail.

I know, all very dodgy like that. The reason I am bringing this up is attestation: We know of many vendors messing things up and even Google rolling our botched attestation extensions. Parsing those must not fail hard, or otherwise no attestation checks are possible.

Now the thing is: We probably don't want something like that for those core extensions that are present, because if those are messed up, the certificate is poisoned anyway. So validating the chain should fail hard. At the same time, it would be nice to know what's in the certificate because as long as an extension is a Triple<OID,Bool?,Asn1Prmitive> it should parse. The was I read this code, it will just throw, so line 87 will hard fail

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
```diff
@@ -96,7 +96,9 @@ class IPAddressName internal constructor(
 
         if (isValid != other.isValid) return false
         if (address != other.address) return false
-        if (addressAndPrefix != other.addressAndPrefix) return false
+//      TODO Add equals for IpInterface in CIDRE
```
JesusMcCloud: Do it!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509CertificateExtension.kt
StjepanovicSrdjan: Wouldn’t a dedicated `InvalidExtension` then be equivalent to current open base class?
What if we instead used an `isValid` pattern, where true indicates that the extension was successfully decoded into one of the known dedicated classes, false indicates that it’s an invalid dedicated extension decoded into the base class, and null means that the OID is unknown but it was successfully decoded into base class?

JesusMcCloud: Not sure. OIDs for extensions are unique, so any contents that do not conform to what is expected for that OID are, by definition, invalid. Again, I am not sure and we should probably discuss this further.

@wkornewald what is your take?

wkornewald: A separate `isValid` property feels like it could be error-prone because people will just forget to do the check and happily accept an invalid extension. For security-critical code I prefer paranoid APIs (same goes for throwing exceptions on signature validation instead of returning true/false because exceptions can't be ignored/forgotten so easily).

StjepanovicSrdjan: Current solution in https://github.com/a-sit-plus/signum/pull/374/commits/53e7643de783bb159dd0dbadf14e8d0dcc5b1c34

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/pkiExtensions/GeneralSubtree.kt
```diff
@@ -0,0 +1,282 @@
...
+                i++
+            }
+        }
+
+        trees = mutableTrees
+        return GeneralSubtrees(mutableTrees)
+    }
+
+    @ExperimentalPkiApi
+    fun unionWith(other: GeneralSubtrees?) {
```
JesusMcCloud: why do we want the ability to union with null?

StjepanovicSrdjan: I looked over it in detail, and yeah — that was definitely an unnecessary mistake that slipped through. Good catch! Fixed in  https://github.com/a-sit-plus/signum/pull/374/commits/1cc85dc454f88f687ef6b39fe240863c145d2bd5

PR #344, opened 2025-08-26
Feature/certificate validation

StjepanovicSrdjan:
Introduce full X.509 certificate validation support
  * Added core `CertificateChainValidator` coordinating the full validation pipeline
  * Modular validator design with pluggable components: 
    * `PolicyValidator` – enforces certificate policies and policy constraints
    * `BasicConstraintsValidator` – validates basicConstraints 
    * `NameConstraintsValidator` – enforces permitted/excluded name constraints across the chain
  * Supports full RFC 5280-style validation rules across name types (DNS, URI, etc.)
* `X509CertificateExtension` is now a base class
  * Enables polymorphic decoding/encoding of extension types
* Refactored `AlternativeNames` for SAN/IAN extraction
  * Removed detailed parsing of individual SAN types; now delegates decoding to `GeneralName`
  
  Clean version of previous PR #297

PR #382, opened 2025-11-28
Feature/certificate validation

StjepanovicSrdjan:
Introduce full X.509 certificate validation support
  * Added core `CertificateChainValidator` coordinating the full validation pipeline
     * `validate()` method returns `CertificateValidationResult` which contains root policy node, leaf certificate and list of `ValidatorFailure`
    * Validation fails softly, by returning `ValidatorFailure` for every exception thrown during validation
  * Modular validator design with pluggable components: 
    * `PolicyValidator` – enforces certificate policies and policy constraints
    * `BasicConstraintsValidator` – validates basicConstraints 
    * `NameConstraintsValidator` – enforces permitted/excluded name constraints across the chain
    * `ChainValidator` - validates signatures and name chaining
    * `KeyUsageValidator` - validates KeyUsage extensions
    * `TimeValidityValidator` - checks certificate time validity and that each certificate was issued within the validity period of its issuer
    * `TrustAnchorValidator` - checks if any certificate from the chain is trusted
  * Supports full RFC 5280-style validation rules across name types (DNS, URI, etc.)

  
  Clean version of previous PR #344 towards next branch

### DIFF
--- a/supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
+++ b/supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
@@ -108,6 +108,7 @@ fun interface ValidatorFactory {
  */
 @OptIn(HazardousMaterials::class)
 @ExperimentalPkiApi
+@Throws(Asn1Exception::class, kotlin.coroutines.cancellation.CancellationException::class)
 suspend fun CertificateChain.validate(
     validatorFactory: ValidatorFactory,
     context: CertificateValidationContext = CertificateValidationContext()
@@ -213,7 +214,7 @@ suspend fun CertificateChain.validate(
  * @throws Asn1Exception on structurally invalid input
  */
 
-@Throws(Asn1Exception::class)
+@Throws(Asn1Exception::class, kotlin.coroutines.cancellation.CancellationException::class)
 @ExperimentalPkiApi
 suspend fun CertificateChain.validate(
     context: CertificateValidationContext = CertificateValidationContext()


indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X509Certificate.kt
```diff
@@ -300,6 +303,50 @@ data class X509Certificate @Throws(IllegalArgumentException::class) constructor(
...
+     */
+    fun isNotYetValid(date: Instant = Clock.System.now()): Boolean =
+        date < tbsCertificate.validFrom.instant
+
+
+    /**
+     * Checks whether this certificate is valid at the specified [date].
+     */
+    @Throws(CertificateValidityException::class)
+    fun checkValidity(date: Instant = Clock.System.now()): Boolean = !(isExpired(date) || isNotYetValid(date))
```
JesusMcCloud: `isValid` or `isValidAt` and no Exception

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,166 @@
...
+    val validator: CertificateValidator? = null,
+    val errorMessage: String,
+    val certificateIndex: Int,
+    val cause: Throwable? = null
+)
+
+/**
+ * Performs a full validation of this [CertificateChain] according to the provided [context].
+ *
+ * EExecutes all registered validators (either provided explicitly
```
JesusMcCloud: Typo

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchor.kt
```diff
@@ -0,0 +1,53 @@
...
+import at.asitplus.signum.indispensable.pki.generalNames.X500Name
+import at.asitplus.signum.indispensable.pki.pkiExtensions.AuthorityKeyIdentifierExtension
+import at.asitplus.signum.supreme.sign.verifierFor
+import at.asitplus.signum.supreme.sign.verify
+
+/**
+ * Represents a trusted certificate authority (TrustAnchor), which can be specified either
+ * as an X509Certificate, or as a distinguished name (in RFC 2253 format) along with a public key.
+ * */
+class TrustAnchor private constructor(
```
JesusMcCloud: Make it a sealed class to distinguish between cert-based and pubkey-based.

Also: is it possible to keep the name optional for pubkey-based? I know not spec compliant, but sometimes you really want a raw public key. It is fine if this requires implementing. a custom validator too.

I'd imagine this to be done using two constructors: One that takes a pub key and a non-nullable name and another hazmat-annotated that takes only a public key without a name

Anyways, this should be discussed offline

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * A [TrustAnchorValidator] is also executed once to ensure that the chain terminates
+ * in a trusted root or intermediate authority.
+ *
+ * @return a [CertificateValidationResult] containing the resulting policy tree,
+ * the end-entity certificate, and a list of any [ValidatorFailure] entries describing validation issues.
+ */
+@ExperimentalPkiApi
+suspend fun CertificateChain.validate(
+    context: CertificateValidationContext = CertificateValidationContext(),
```
JesusMcCloud: I'd to this differently: Create two functions: one that takes no params at all and one that takes a list

- The no-param function calls the other and passes the default context containing everything. Maybe even add a function with a self-explanatory name that constructs this context, such as `newRFC5280CertificateValidationContext()` or something like that.
- The second function that takes the context gets a HazMat annotation, so that by default, you will get the RFC5280-compliant validation, but, iff you opt into HazMat, you can have free reign and even NOT add any of the mandatory validators. This way you can support custom foo that is a bit more permissive, for example für Android Attestation cert path validation, as that chain is not RFC5280-compliant.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,166 @@
...
+    validators.addIfMissing(KeyUsageValidator(this.size, expectedEku = context.expectedEku))
+    validators.addIfMissing(BasicConstraintsValidator(this.size))
+    validators.addIfMissing(ChainValidator(this.reversed()))
+    validators.addIfMissing(TimeValidityValidator(context.date, certificateChain = this.reversed()))
+
+    val activeValidators = validators.toMutableSet()
+    val validatorFailures = emptyList<ValidatorFailure>().toMutableList()
+    val trustAnchorValidator = TrustAnchorValidator(context.trustAnchors, this)
+
+    try {
```
JesusMcCloud: Use `catchingUnwrapped {…}.onFialure{…}`, because fatal exceptions then still get thrown

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
StjepanovicSrdjan: I agree with this approach, and I implemented it with small adjustments in https://github.com/a-sit-plus/signum/pull/382/commits/53947e32d0b82e176c3c6162281b022b83d782e0. 
Now we have a separate `validate()` function, but the default one that follows RFC 5280 still requires a context so it can construct the default validators (policy flags, validation time, etc.). I excluded the validators list from the context. Based on the context, the RFC 5280 version constructs the default validators, then calls the second hazardous validate() function, which accepts a list of validators.

This hazardous function can also be used for custom validation, where the user can construct a custom list of validators and pass it to `validate()`.

Additionally, I added defineRFC5280Validators(), which returns a `MutableList` of `CertificateValidator`. This allows the user to get the default validators, then remove unnecessary ones or add custom ones.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchor.kt
StjepanovicSrdjan: Done in https://github.com/a-sit-plus/signum/pull/382/commits/99cf164c56f35e4ca694b124bc008f80f30c80d4. I made the name optional, even though it is not spec compliant, it should still be safe, because validation will fail if names are not equal in `isIssuerOf()` method.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -185,14 +189,21 @@ private fun defineRFC5280Validators(
     chain: CertificateChain
 ): MutableList<CertificateValidator> {
     val validators = mutableListOf<CertificateValidator>()
+    val (trustAnchors, pathLen, processingChain) =
+        if (context.allowIncludedTrustAnchor) {
+            val newAnchors = setOf(TrustAnchor.Certificate(chain.root)) + context.trustAnchors
+            Triple(newAnchors, chain.size - 1, chain.dropLast(1))
```
JesusMcCloud: But then this is not allow, but enforce? After all, this always drops the last cert, when it should really check, if the root from the chain is part of the trust anchors and only then remove it from the chain. Our Cert class handles equality checks, so this should be smooth. Also: dedicated tests for this logic are missing and we want to only treat the root as trusted if it really is part of the specified trust anchors.

StjepanovicSrdjan: done in https://github.com/a-sit-plus/signum/pull/382/commits/1475dfc353268e99fce8bca6a98bf3e0bf92afd5

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/TimeValidityValidator.kt
```diff
@@ -17,33 +17,42 @@ import kotlin.time.Instant
...
+        if (currentCertIndex < certChain.lastIndex - 1 || (currentCertIndex == certChain.lastIndex - 1 && checkLeafValidity)) {
             val childCert = certChain[currentCertIndex + 1]
             wasCertificateIssuedWithinIssuerValidityPeriod(
                 dateOfIssuance = childCert.tbsCertificate.validFrom.instant,
-                issuer = currCert)
-            currentCertIndex++
+                issuer = currCert
+            )
         }
+        currentCertIndex++
```
JesusMcCloud: I'd add this more complex variant to the test sources only. because it is non-standard and we don't need that extra complexity as part of the main sources. We can then lift into the main sources as a distinct TimeValitityValidator variant for attestation cert chain checks

StjepanovicSrdjan: done in https://github.com/a-sit-plus/signum/pull/382/commits/762bc62f7c18f45be4beec23fcfd04a80ce1da4f

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/BasicConstraintsValidator.kt
```diff
@@ -0,0 +1,58 @@
...
+    private var remainingPathLength: UInt? = null,
+    private var currentCertIndex: Int = 0
+) : CertificateValidator {
+
+    @ExperimentalPkiApi
+    override suspend fun check(currCert: X509Certificate, remainingCriticalExtensions: MutableSet<ObjectIdentifier>) {
+        remainingCriticalExtensions.remove(KnownOIDs.basicConstraints_2_5_29_19)
+        if (currentCertIndex >= certPathLen - 1) return
+
+        currentCertIndex++
```
wkornewald, line 24: Maybe it's too late to comment in this PR since it's closed already, but we really should make validator instances immutable and not depend on this function getting called multiple times in the right order. This is way too fragile and there's too much potential for mistakenly creating security issues. Instead, if something needs to look into the whole chain then let's better introduce a separate CertificateChainValidator interface for global checks which go beyond a single cert. Or if there should be just one single validation API then let's rethink the whole concept.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/CertificateValidator.kt
```diff
@@ -0,0 +1,14 @@
...
+import at.asitplus.signum.CertificateException
+import at.asitplus.signum.ExperimentalPkiApi
+import at.asitplus.signum.indispensable.asn1.ObjectIdentifier
+import at.asitplus.signum.indispensable.pki.X509Certificate
+import kotlin.coroutines.cancellation.CancellationException
+
+interface CertificateValidator {
+    // Every validator removes checked critical extensions
+    @ExperimentalPkiApi
+    @Throws(CertificateException::class, CancellationException::class)
```
wkornewald, line 12: If this is supposed to be callable from Swift/Obj-C and any other exception type (e.g. IllegalStateException) could in theory be thrown then this would hard-crash the whole app. While it's nice to document the exceptions that a function is supposed to throw, I prefer to put that in the docstring and always use `@Throws(Throwable::class)` in order to prevent hard-crashing iOS apps.

BTW, I also maintain a compiler plugin which simplifies this because Kotlin's behavior has bitten us way too often https://github.com/ensody/kotlin-default-throws-plugin

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/CertificateValidator.kt
```diff
@@ -0,0 +1,14 @@
...
+    suspend fun check(currCert: X509Certificate, remainingCriticalExtensions: MutableSet<ObjectIdentifier>)
```
wkornewald, line 13: The remainingCriticalExtensions parameter feels backwards. Shouldn't each validator just add the extensions that it has checked instead of removing them? This information could even be helpful for debugging.

Also, instead of mutating the set of extensions (which could allow invalid manipulations), each validator could just return what it has checked. This way everything becomes immutable and invalid manipulations become impossible.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,285 @@
...
+            rootNode = PolicyNode(
+                parent = null,
+                validPolicy = KnownOIDs.anyPolicy,
+                criticalityIndicator = false,
+                expectedPolicySet = setOf(KnownOIDs.anyPolicy),
+                generatedByPolicyMapping = false
+            )
+        )
+    )
+    validators += listOf(
```
wkornewald, line 253: You can just return everything in a single listOf(). There is no need for a mutableListOf and the validators.add() call above is also unnecessary.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/CertificateValidator.kt
JesusMcCloud: I am always for immutability when possible!

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/BasicConstraintsValidator.kt
JesusMcCloud: Not too late, as everything's experimental and we didn't even get to really dogfood this before Christmas.
Two distinct interfaces might be one way to go. Then again, I've really only come across three cases that deviate from the default cert chain validation logic and implementing those using the API as it stands now is already pretty smooth. From what I gathered, the current way of doing things has the advantage of making it possible to implement cert path validation specs very close to how they are phrased in natural language.

Not saying that things need to stay, as two distinct interfaces might do away with *most* mutability, if not all, so I like the idea. @StjepanovicSrdjan what's your take?

wkornewald: BTW, one reason for making validators immutable is that it's safe to define the list of validators just once. For example:

```kotlin
TrustStore(validators = listOf(...))
```

With the current implementation you'd have to be really careful to never use a validator for more than one validation of a single chain and all APIs would instead have to take a lambda returning fresh instances (`validators = { listOf(...) }`). The risk of making a mistake is just unnecessarily high.

Also, an error in a single cert of the chain potentially makes the validator unsafe to use for the rest of the chain because there's no guarantee that the mutated state is fully prepared for the next call. So, there's no safe way to validate all the certs and collect multiple errors.

And the single CertValidator is just a special case of a CertChainValidator:

```kotlin
interface CertValidator : CertChainValidator {
    fun check(cert: X509Certificate)

    override fun check(chain: CertificateChain) {
        chain.forEach { check(it) }
    }
}
```

So you actually need just one single type (CertChainValidator). The CertValidator is a just helper API for the simple single-cert case.

I've intentionally left out the return values in that code sample, but those functions would have to return the set of checked extensions. And in case of the cert chain validator that would have to be a map from cert to checked extensions. But the details of how you want to model the types are unrelated to the general principle.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,285 @@
...
+        nameConstraintsValidator?.previousNameConstraints = trustAnchorValidator.trustAnchor?.nameConstraints
+        activeValidators.remove(trustAnchorValidator)
+    }
+
+    processingChain.reversed().forEachIndexed { i, cert ->
+        val remainingCriticalExtensions = cert.criticalExtensionOids.toMutableSet()
+
+        val validatorIterator = activeValidators.iterator()
+        while (validatorIterator.hasNext()) {
+            val currValidator = validatorIterator.next()
```
wkornewald, line 169: Can't you just use a simple for-loop?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/TimeValidityValidator.kt
```diff
@@ -0,0 +1,59 @@
...
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+import kotlin.time.Instant
+
+/**
+ * Checks the validity of the each certificate in the chain based on the given date and
+ * confirms that each certificate was issued within the validity period of its issuer
+ */
+class TimeValidityValidator(
+    val date: Instant,
```
wkornewald, line 18: It would be better to provide a Clock instead of a fixed Instant, so this validator doesn't have to be re-instantiated for every validation. The Clock implementation itself could still return a fixed Instant where that is desired (e.g. in unit tests).

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/CertValidityValidator.kt
```diff
@@ -0,0 +1,53 @@
...
+import at.asitplus.signum.indispensable.pki.pkiExtensions.SubjectKeyIdentifierExtension
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+import kotlin.time.Instant
+
+/**
+ * Checks whether the certificate is constructed correctly, since some components are decoded too leniently
+ * */
+class CertValidityValidator(
+    val date: Instant
```
wkornewald, line 21: This value is unused.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/KeyUsageValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+import at.asitplus.signum.indispensable.pki.pkiExtensions.KeyUsageExtension
+
+/**
+ * Ensures that intermediate CA certificates have the necessary key usage extensions.
+ * Key usage to sign certificates (keyCertSign) and sign certificate revocation lists (cRLSign), according to RFC 5280.
+ */
+class KeyUsageValidator (
+    private val certPathLen: Int,
+    private var currentCertIndex: Int = 0,
+    private val expectedEku: Set<ObjectIdentifier> = emptySet()
```
wkornewald, line 21: It would be nice to emphasize somehow that this is the leaf cert's expected EKU only. Possibly the EKU validator should actually be a separate class, so you can use that stand-alone without doing the other checks unnecessarily.

Also, support for requiring additional leaf cert key usages would be nice.
And maybe sometimes the cRLSign check should be optional when used outside of the RFC 5280 context.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/KeyUsageValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+) : CertificateValidator {
+
+    private var supportedExtensions: Set<ObjectIdentifier> = setOf(
```
wkornewald, line 24: This shouldn't be a var, but a val.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/KeyUsageValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+        KnownOIDs.keyUsage,
+        KnownOIDs.extKeyUsage,
+        KnownOIDs.subjectAltName_2_5_29_17,
```
wkornewald, line 27: I don't think the SAN is actually checked by this validator. The CertValidityValidator takes care of this only.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/KeyUsageValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+    )
+
+    @ExperimentalPkiApi
+    override suspend fun check(currCert: X509Certificate, remainingCriticalExtensions: MutableSet<ObjectIdentifier>) {
+        remainingCriticalExtensions.removeAll(supportedExtensions)
```
wkornewald, line 32: It's not quite correct to mark all those extensions as handled. The leaf cert's KeyUsageExtension is never looked at unless it's a CA cert. Maybe if there was an additional config option for the leaf's expected extension (or a lambda to allow arbitrary checks) then it could be marked as handled.

The way the code works now it would mistakenly mark the leaf cert's KeyUsageExtension as handled although we need to install a separate validator and now it's too easy to forget.

indispensable/src/commonMain/kotlin/at/asitplus/signum/Throwables.kt
```diff
@@ -3,3 +3,12 @@ package at.asitplus.signum
 sealed class CryptoException(message: String? = null, cause: Throwable? = null) : Throwable(message, cause)
 open class CryptoOperationFailed(message: String) : CryptoException(message)
 open class UnsupportedCryptoException(message: String? = null, cause: Throwable? = null) : CryptoException(message, cause)
+
+sealed class CertificateException(message: String? = null, cause: Throwable? = null) : Throwable(message, cause)
```
wkornewald, line 7: Would it be possible to make this non-sealed, so custom validators can extend this?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchorValidator.kt
```diff
@@ -0,0 +1,87 @@
...
+class TrustAnchorValidator(
+    private val trustAnchors: Set<TrustAnchor>,
+    private val certChain: CertificateChain,
+    private var currentCertIndex: Int = 0,
+    var trustAnchor: TrustAnchor? = null,
+    val date: Instant
+) : CertificateValidator {
+
+    var foundTrusted: Boolean = false
+    private val basicConstraintsValidator: BasicConstraintsValidator = BasicConstraintsValidator(0)
```
wkornewald, line 28: This leads to incorrect error messages because the index is always 0. It's better to make checkCaBasicConstraints a reusable function outside of the BasicConstraintsValidator.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchor.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    suspend fun isIssuerOf(cert: X509Certificate): Boolean {
+        val verifier = (cert.signatureAlgorithm as X509SignatureAlgorithm).verifierFor(publicKey).getOrElse { return false }
+        val signatureValid = verifier.verify(
+            cert.tbsCertificate.encodeToDer(),
+            cert.decodedSignature.getOrThrow()
+        ).isSuccess
+
+        val issuerName = cert.tbsCertificate.issuerName
+        return signatureValid && issuerName == principal
```
wkornewald, line 60: Shouldn't this also check the issuerUniqueID? At least the ChainValidator does this and this check is inconsistent.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchor.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun matchesCertificate(cert: X509Certificate): Boolean {
+        this.cert?.let { return it == cert }
+
+        val sameKey = runCatching { this.publicKey == cert.decodedPublicKey.getOrThrow() }.getOrElse { false }
+        val samePrincipal = this.principal == cert.tbsCertificate.subjectName
+        return sameKey && samePrincipal
+    }
+
+    suspend fun isIssuerOf(cert: X509Certificate): Boolean {
+        val verifier = (cert.signatureAlgorithm as X509SignatureAlgorithm).verifierFor(publicKey).getOrElse { return false }
```
wkornewald, line 53: The signature check should happen last because it's expensive.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/validate/TimeValidityValidator.kt
```diff
@@ -0,0 +1,59 @@
...
+) : CertificateValidator {
+
+    @ExperimentalPkiApi
+    override suspend fun check(
+        currCert: X509Certificate,
+        remainingCriticalExtensions: MutableSet<ObjectIdentifier>
+    ) {
+        if (currCert.isExpired(date)) {
+            throw CertificateValidityException(
+                "certificate expired on " + currCert.tbsCertificate.validUntil.instant.toLocalDateTime(TimeZone.UTC)
```
wkornewald, line 30: Why not TimeZone.currentSystemDefault()? Also for the check below.

Wouldn't it make sense to define a helper function X509Certificate.verifyValidAt(instant), so you don't have to duplicate the same code in TrustAnchorValidator?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchorValidator.kt
```diff
@@ -0,0 +1,87 @@
...
+import at.asitplus.signum.CertificateValidityException
+import at.asitplus.signum.ExperimentalPkiApi
+import at.asitplus.signum.indispensable.asn1.ObjectIdentifier
+import at.asitplus.signum.indispensable.pki.CertificateChain
+import at.asitplus.signum.indispensable.pki.X509Certificate
+import at.asitplus.signum.indispensable.pki.validate.BasicConstraintsValidator
+import at.asitplus.signum.indispensable.pki.validate.CertificateValidator
+import kotlinx.datetime.TimeZone
+import kotlinx.datetime.toLocalDateTime
+import org.kotlincrypto.error.CertificateException
```
wkornewald, line 12: The import is wrong. You want the Signum class instead.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchorValidator.kt
```diff
@@ -0,0 +1,87 @@
...
+                        "certificate expired on " + currCert.tbsCertificate.validUntil.instant.toLocalDateTime(
+                            TimeZone.currentSystemDefault()
+                        )
+                    )
+                }
+            }
+
+            issuingAnchor.cert?.let {
+                if (it.isNotYetValid(date)) {
+                    throw CertificateValidityException(
```
wkornewald, line 71: Could you please define more specific exception classes for each validation, so it's possible to distinguish them in detail? Sometimes it's required e.g. for a backend to return very specific error codes for each individual validation failure.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,285 @@
...
+ *
+ * @return a [CertificateValidationResult] containing the resulting policy tree and
+ * the end-entity certificate, or a list of [ValidatorFailure]s for a chain that failed to validate.
+ *
+ * @throws Asn1Exception on structurally invalid input
+ */
+@OptIn(HazardousMaterials::class)
+@ExperimentalPkiApi
+@Throws(Asn1Exception::class, kotlin.coroutines.cancellation.CancellationException::class)
+suspend fun CertificateChain.validate(
```
wkornewald, line 112: The whole logic in this function is way too complicated and error-prone.

* Why do we manually call `keyIdentifierValidator?.checkTrustAnchorAndChild()` within special-case code for `trustAnchorValidator`?
  * Why isn't that check part of `TurstAnchorValidator`? The code could just be a global reusable helper function instead of living in `KeyIdentifierValidator`.
  * If `checkTrustAnchorAndChild()` fails, the `KeyIdentifierValidator.currentCertIndex` variable will be always 0 at that point and thus won't even match `TurstAnchorValidator.currentCertIndex`. Such mutation based code becomes messy very quickly.
* Why is the `trustAnchorValidator` using `processingChain` and the other validators `processingChain.reversed()`?
* Why do we need messy mutations like `nameConstraintsValidator?.previousNameConstraints = trustAnchorValidator.trustAnchor?.nameConstraints`?

Once the cert validators work on a whole chain instead of individual certs some of the complexity might go away, but still: If different validators need to communicate with each other then they should better be merged into one single validator that contains the whole logic without any mutation. Validators should be stand-alone usable and ideally the validation should be just as simple as `val checkedExtensions = validators.map { it.check(chain) }; checkUnknownCrtiticalExtensions(chain, checkedExtensions.reduce());`. They should never require some outer caller to coordinate with complicated extra logic.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
StjepanovicSrdjan: Not exactly, a simple for loop won’t work here because we remove validators that fail.

wkornewald: You could loop over the iterator at least. Then you don't need the hasNext() and next() logic, but you can still call validatorIterator.remove().

StjepanovicSrdjan: I believe for loop still doesn't handle removing properly, it throws `ConcurrentModificationException`

wkornewald: When looping over a mutable iterator it should work. I've tested it.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TrustAnchorValidator.kt
```diff
@@ -0,0 +1,87 @@
...
+    override suspend fun check(
+        currCert: X509Certificate,
+        remainingCriticalExtensions: MutableSet<ObjectIdentifier>
+    ) {
+        if (foundTrusted) return
+        val issuingAnchor = trustAnchors.firstOrNull { anchor ->
+            anchor.isIssuerOf(currCert)
+        }
+
+        if (issuingAnchor != null) {
```
wkornewald, line 40: Maybe I haven't look right, but this validator seems to only check if any cert in the chain is signed by a trust anchor. However, where do we check that each cert in the chain is really signed by the next cert? The only usage of TrustAnchor.isIssuerOf() in the whole project is in the firstOrNull above.

So, I could construct a chain a, b where a is an arbitrary self-signed MITM cert and b is a valid Let's Encrypt intermediate issued by a trust anchor, but a is not actually signed by b. Does some something in the code prevent this?

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/CertificateChainValidator.kt
```diff
@@ -0,0 +1,285 @@
...
+    validatorFactory: ValidatorFactory,
+    context: CertificateValidationContext = CertificateValidationContext()
+): CertificateValidationResult {
+
+    val validators = with(validatorFactory) { this@validate.generate(context) }
+    val processingChain = if (context.allowIncludedTrustAnchor && context.trustAnchors.any {
+            it.matchesCertificate(this.root)
+        }) this.dropLast(1) else this
+
```
wkornewald, line 121: Instead of this logic and chains being inconsistent, it would be good to introduce a chain constructor interface which can optionally be implemented by the validators to provide a Sequence of candidate chains to check (supporting cross-signing). The TrustAnchorValidator could implement this.

Also, the candidate chains should always include the trust anchor in the end, so e.g. time validity and other validations always apply to the trust anchor, too. Otherwise it's too easy to forget important checks.

PR #399, opened 2026-01-16
Work around KTLC-358

kyay10:
Everything seems to compile and run fine!


kyay10:
The verify CLA job needs to be re-run

PR #401, opened 2026-01-21
Redesign Symmetric types

kyay10:
The key trick is to use `SymmetricEncryptionAlgorithm` to hold the type parameters (i.e. Auth and Nonce). Then, the sealed hierarchies no longer rely on interfaces, instead, they provide extensions that `when` over the relevant inheritors, and use `absurd` when an inheritor cannot possibly be called on this method.


kyay10:
There's one test in `ValidationApiTest` that seems broken due to a certificate expiry. I get the failure without my changes, so it's not my code that caused it lol

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/generalNames/IPAddressName.kt
JesusMcCloud, line 1: This is unrelated to symmetric crypto

PR #419, opened 2026-03-11
Full implementation of all three JWS serialization shapes

n0900:
Implements three new classes `JwsCompact`, `JwsFlattened` and `JwsGeneral` to implement RFC 7515 Section 3.1, 7.2.1 and 7.2.2 respectively.

Removed generic parameter as present in `JwsSigned` which will be deprecated. Payload can still be serialized into a specific structure at runtime. Serialization stability is now guaranteed by handling all signature related parameters as bytearrays and deserializing into transient properties only.

### DIFF
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -1,0 +1,44 @@
+package at.asitplus.signum.indispensable.josef
+
+import at.asitplus.signum.indispensable.contentEqualsIfArray
+import at.asitplus.signum.indispensable.io.ByteArrayBase64UrlSerializer
+import at.asitplus.signum.indispensable.josef.io.joseCompliantSerializer
+import kotlinx.serialization.*
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+
+sealed class JWS {
+    abstract val payload: ByteArray
+
+    fun <P> getPayload(serializer: KSerializer<P>, serialFormat: SerialFormat = joseCompliantSerializer): P =
+        when (serialFormat) {
+            is StringFormat -> serialFormat.decodeFromString(serializer, payload.decodeToString())
+            is BinaryFormat -> serialFormat.decodeFromByteArray(serializer, payload)
+            else -> throw NotImplementedError("Unknown serial format $serialFormat")
+        }
+
+    /**
+     * Find correct serializer at compile time
+     */
+    inline fun <reified P> getPayload(serialFormat: SerialFormat): P =
+        getPayload(serialFormat.serializersModule.serializer(), serialFormat)
+
+    object SerialNames {
+        const val PROTECTED = "protected"
+        const val HEADER = "header"
+        const val SIGNATURE = "signature"
+        const val SIGNATURES = "signatures"
+    }
+}
+
+internal fun JsonObject?.strictUnion(other: JsonObject?): JsonObject {
+    if (this == null) return other ?: JsonObject(emptyMap())
+    if (other == null) return this
+
+    val duplicates = this.keys intersect other.keys
+    require(duplicates.isEmpty()) {
+        "Duplicate keys: ${duplicates.joinToString()}"
+    }
+
+    return JsonObject(this + other)
+}
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -1,12 +1,17 @@
 package at.asitplus.signum.indispensable.josef
 
-import at.asitplus.signum.indispensable.contentEqualsIfArray
-import at.asitplus.signum.indispensable.io.ByteArrayBase64UrlSerializer
+import at.asitplus.signum.indispensable.CryptoSignature
+import at.asitplus.signum.indispensable.io.Base64UrlStrict
 import at.asitplus.signum.indispensable.josef.io.joseCompliantSerializer
+import io.matthewnelson.encoding.core.Encoder.Companion.encodeToString
 import kotlinx.serialization.*
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.decodeFromJsonElement
 
+
+/**
+ * Wrapper for all JWS formats.
+ */
 sealed class JWS {
     abstract val payload: ByteArray
 
@@ -23,11 +28,32 @@ sealed class JWS {
     inline fun <reified P> getPayload(serialFormat: SerialFormat): P =
         getPayload(serialFormat.serializersModule.serializer(), serialFormat)
 
+    /**
+     * Lenient Signature Parsing
+     */
+    fun getSignature(combinedHeader: JwsHeader, plainSignature: ByteArray): CryptoSignature.RawByteEncodable =
+        when (val alg = combinedHeader.algorithm) {
+            is JwsAlgorithm.Signature.EC -> CryptoSignature.EC.fromRawBytes(alg.ecCurve, plainSignature)
+            is JwsAlgorithm.Signature.RSA -> CryptoSignature.RSA(plainSignature)
+            else -> throw SerializationException("Unsupported algorithm for JWS signature element: $alg")
+        }
+
+    fun getCombinedHeader(unprotectedHeader: JsonObject, protectedHeader: ByteArray): JwsHeader =
+        joseCompliantSerializer.decodeFromJsonElement(
+            unprotectedHeader.strictUnion(
+                joseCompliantSerializer.decodeFromString(protectedHeader.decodeToString())
+            )
+        )
+
+    fun getSignatureInput(protectedHeader: ByteArray, payload: ByteArray) =
+        "${protectedHeader.encodeToString(Base64UrlStrict)}.${payload.encodeToString(Base64UrlStrict)}".encodeToByteArray()
+
     object SerialNames {
         const val PROTECTED = "protected"
         const val HEADER = "header"
         const val SIGNATURE = "signature"
         const val SIGNATURES = "signatures"
+        const val PAYLOAD = "payload"
     }
 }
 
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -42,6 +42,12 @@ sealed class JWS {
         const val SIGNATURE = "signature"
         const val SIGNATURES = "signatures"
         const val PAYLOAD = "payload"
+
+        /* Shapes */
+        const val COMPACT = "compact"
+        const val FLATTENED = "flattened"
+        const val GENERAL = "general"
+
     }
 
     companion object {
@@ -62,9 +68,9 @@ sealed class JWS {
     object JwsSerializer: KSerializer<JWS> {
         @OptIn(InternalSerializationApi::class)
         override val descriptor: SerialDescriptor = buildSerialDescriptor("JWS", PolymorphicKind.SEALED) {
-            element("compact", JwsCompactStringSerializer.descriptor)
-            element("flattened", JwsFlattened.serializer().descriptor)
-            element("general", JwsGeneral.serializer().descriptor)
+            element(SerialNames.COMPACT, JwsCompactStringSerializer.descriptor)
+            element(SerialNames.FLATTENED, JwsFlattened.serializer().descriptor)
+            element(SerialNames.GENERAL, JwsGeneral.serializer().descriptor)
         }
 
         override fun serialize(
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -20,6 +20,8 @@ import kotlinx.serialization.json.JsonPrimitive
 
 /**
  * Wrapper for all JWS formats.
+ *
+ * If [plainPayload] data structure is defined as part of the contact consider [JwsTyped]
  */
 @Serializable(with = JWS.JwsSerializer::class)
 sealed class JWS {
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -61,7 +61,7 @@ sealed class JWS {
     object JwsSerializer: KSerializer<JWS> {
         @OptIn(InternalSerializationApi::class)
         override val descriptor: SerialDescriptor = buildSerialDescriptor("JWS", PolymorphicKind.SEALED) {
-            element("compact", JwsCompact.serializer().descriptor)
+            element("compact", JwsCompactStringSerializer.descriptor)
             element("flattened", JwsFlattened.serializer().descriptor)
             element("general", JwsGeneral.serializer().descriptor)
         }
@@ -71,7 +71,7 @@ sealed class JWS {
             value: JWS
         ) {
             when (value) {
-                is JwsCompact -> encoder.encodeSerializableValue(JwsCompact.serializer(), value)
+                is JwsCompact -> encoder.encodeSerializableValue(JwsCompactStringSerializer, value)
                 is JwsFlattened -> encoder.encodeSerializableValue(JwsFlattened.serializer(), value)
                 is JwsGeneral -> encoder.encodeSerializableValue(JwsGeneral.serializer(), value)
             }
@@ -82,7 +82,7 @@ sealed class JWS {
             val jsonElement = decoder.decodeJsonElement()
 
             return when (jsonElement) {
-                is JsonPrimitive -> decoder.json.decodeFromJsonElement(JwsCompact.serializer(), jsonElement)
+                is JsonPrimitive -> decoder.json.decodeFromJsonElement(JwsCompactStringSerializer, jsonElement)
                 is JsonObject -> {
                     val hasGeneralSignatures = SerialNames.SIGNATURES in jsonElement
                     val hasFlattenedSignature = SerialNames.SIGNATURE in jsonElement
--- a/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
+++ b/indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
@@ -13,7 +13,6 @@ import kotlinx.serialization.encoding.Encoder
 import kotlinx.serialization.json.JsonDecoder
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
-import kotlinx.serialization.json.decodeFromJsonElement
 
 
 /**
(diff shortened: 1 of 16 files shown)


indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsSigned.kt
```diff
@@ -20,6 +20,7 @@ import kotlinx.serialization.json.Json
  *
  * See [RFC 7515](https://datatracker.ietf.org/doc/html/rfc7515)
  */
+@Deprecated("", level = DeprecationLevel.WARNING)
```
nodh: Should add a message here on how to replace usages of this class

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+ * BASE64URL(UTF8(HEADER)).BASE64URL(PAYLOAD).BASE64URL(SIGNATURE)
+ *
+ * This class does not support an unprotected header field!
+ *
+ * When [JwsCompact] is serialized through a JSON format, the compact representation becomes a JSON string literal.
+ * That is correct for nested usage inside a larger JSON document, but the surrounding JSON text contains quotes.
+ *
+ * For a standalone compact JWS string, use [toString] and [JwsCompact.invoke]
+ */
+@Serializable(with = JwsCompact.JwsCompactSerializer::class)
```
JesusMcCloud: I stand by my conviction. This should not be not serializable

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,127 @@
...
+                    "Invalid JWS compact serialization: expected 3 parts, got ${parts.size}"
+                )
+            }
+
+            return try {
+                JwsCompact(
+                    plainProtectedHeader = parts[0].decodeToByteArray(Base64UrlStrict),
+                    payload = parts[1].decodeToByteArray(Base64UrlStrict),
+                    plainSignature = parts[2].decodeToByteArray(Base64UrlStrict),
+                )
```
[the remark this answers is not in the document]
JesusMcCloud: silicone devil is wrong. eager init of the header proerty takes care of that

JesusMcCloud (changes requested):
should not be serailiazable. Otherwise, I like the semantics and how it is constructed, so keep the class but get rid of it being serializable

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
nodh: I for one find it useful to have it serializable, because we're gonna use it that way it some classes.

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/Jws.kt
```diff
@@ -0,0 +1,129 @@
...
+            protectedHeader?.encodeToString(Base64UrlStrict).orEmpty()
+
+        fun getSignatureInput(protectedHeader: ByteArray?, payload: ByteArray) =
+            "${getEncodedProtectedHeader(protectedHeader)}.${payload.encodeToString(Base64UrlStrict)}".encodeToByteArray()
+    }
+
+    object JwsSerializer: KSerializer<JWS> {
+        @OptIn(InternalSerializationApi::class)
+        override val descriptor: SerialDescriptor = buildSerialDescriptor("JWS", PolymorphicKind.SEALED) {
+            element("compact", JwsCompact.serializer().descriptor)
```
nodh: could be extracted into the `SerialNames` I guess

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+ * This class does not support an unprotected header field!
+ *
+ * When [JwsCompact] is serialized through a JSON format, the compact representation becomes a JSON string literal.
+ * That is correct for nested usage inside a larger JSON document, but the surrounding JSON text contains quotes.
+ *
+ * For a standalone compact JWS string, use [toString] and [JwsCompact.invoke]
+ */
+@Serializable(with = JwsCompact.JwsCompactSerializer::class)
+data class JwsCompact(
+    @Serializable(ByteArrayBase64UrlSerializer::class)
```
nodh: Is this relevant when the class itself is marked with a custom serializer?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+    override val payload: ByteArray,
+    @Serializable(ByteArrayBase64UrlSerializer::class)
+    val plainSignature: ByteArray,
+) : JWS() {
+
+    @Transient
+    val jwsHeader = JwsHeader.fromParts(plainProtectedHeader, null)
+
+    val signature by lazy { getSignature(jwsHeader.algorithm, plainSignature) }
```
nodh: Should be transient property to force eager parsing

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+    val signatureInput by lazy { getSignatureInput(plainProtectedHeader, payload) }
```
nodh: Should be transient property to force eager parsing

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    override fun toString(): String {
+        val signingInput =
```
nodh: looks like this would easily fit into one line with 120 chars

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+            getSignatureInput(plainProtectedHeader, payload).decodeToString()
+        val signature =
```
nodh: looks like this would easily fit into one line with 120 chars

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+                throw SerializationException("Invalid base64url content in JWS compact serialization", e)
+            }
+        }
+
+        operator fun invoke(
+            protectedHeader: JwsHeader,
+            payload: ByteArray,
+            signer: (JwsAlgorithm, ByteArray) -> ByteArray
+        ): JwsCompact {
+            val plainProtectedHeader = JwsProtectedHeaderSerializer.encodeToByteArray(protectedHeader.toPart())
```
nodh, line 121: variable might be inlined and then the function can be converted to an expression body

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,129 @@
...
+            return JwsCompact(
+                plainProtectedHeader = plainProtectedHeader,
+                payload = payload,
+                plainSignature = signer(protectedHeader.algorithm, getSignatureInput(plainProtectedHeader, payload)),
+            )
+        }
+    }
+}
+
+fun JwsCompact.toJwsFlattened(): JwsFlattened = JwsFlattened(
```
nodh, line 150: why is that an extension function and not a proper function inside the class?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,101 @@
...
+    override val payload: ByteArray,
+    @Serializable(ByteArrayBase64UrlSerializer::class)
+    @SerialName(SerialNames.SIGNATURE)
+    val plainSignature: ByteArray
+) : JWS() {
+
+    @Transient
+    val jwsHeader = JwsHeader.fromParts(plainProtectedHeader, unprotectedHeader)
+
+    val signature by lazy { getSignature(jwsHeader.algorithm, plainSignature) }
```
nodh: Should be a transient property to force eager evaluation

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    companion object {
+        operator fun invoke(
+            protectedHeader: JwsHeader.Part?,
+            unprotectedHeader: JwsHeader.Part?,
+            payload: ByteArray,
+            signer: (JwsAlgorithm, ByteArray) -> ByteArray
+        ): JwsFlattened {
+            val jwsHeader = JwsHeader.fromParts(protectedHeader, unprotectedHeader)
```
nodh: Variables might be inlined and then the function can be an expression body

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,101 @@
...
+            )
+        }
+    }
+}
+
+fun JwsFlattened.toJwsCompact(): JwsCompact {
+    require(unprotectedHeader == null) {"Compact Serialization does not support unprotected header"}
+    runCatching { JwsHeader.fromParts(plainProtectedHeader) }.getOrElse { throw IllegalArgumentException("Compact JWS requires protected header to be a valid JwsHeader") }
+    return JwsCompact(
+        plainProtectedHeader = requireNotNull(plainProtectedHeader) {
```
nodh: All requirements should be handled the same way: Either before the return statement or inline

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsHeader.kt
```diff
@@ -7,15 +7,19 @@ import at.asitplus.signum.indispensable.CryptoPublicKey
...
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.UseSerializers
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.jsonObject
+import kotlin.time.Instant
 
 /**
  * Header of a [JwsSigned].
```
nodh: Comment isn't accurate anymore

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsHeader.kt
```diff
@@ -300,14 +406,42 @@ data class JwsHeader(
...
+        internal fun fromJsonObjects(
+            protectedHeader: JsonObject? = null,
+            unprotectedHeader: JsonObject? = null,
+        ): JwsHeader = joseCompliantSerializer.decodeFromJsonElement<JwsHeader>(
+            protectedHeader.strictUnion(unprotectedHeader)
+        )
+
         // TODO usages!
         @Deprecated("To be removed in next release")
         fun deserialize(it: String) = catching {
```
nodh: Is it time to remove that function now?

indispensable-josef/src/commonTest/kotlin/at/asitplus/signum/indispensable/josef/JwsSerializationTest.kt
```diff
@@ -0,0 +1,326 @@
...
+        val missingHeaderResult = runCatching { missingProtectedHeader.toJwsCompact() }
+        val missingPartResult = runCatching { JwsCompact("a.b") }
+        val extraPartResult = runCatching { JwsCompact("a.b.c.d") }
+        val invalidBase64Result = runCatching { JwsCompact("!!.e30.AQ") }
+
+        missingHeaderResult.isSuccess shouldBe false
+        missingHeaderResult.exceptionOrNull().shouldBeInstanceOf<IllegalArgumentException>()
+
+        missingPartResult.isSuccess shouldBe false
+        missingPartResult.exceptionOrNull()?.message?.shouldContain("expected 3 parts, got 2")
```
nodh: Better to have a `.shouldNotBeNull()` here to avoid the optional chaining ... or maybe even a `shouldThrow` above?

nodh (changes requested):
I like it, some minor remarks

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
JesusMcCloud: it is. base64url is not the default way of serializing bytes.

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
JesusMcCloud: but it is not valid json, it is just a string and we tripped over this footgun in 2024 already. so how would we keep others from tripping over it? maybe keep the serializer, but do not annotate the class as serializable, but document that the serializer is present. maybe even call it `JwsCompactSTringSerializer` and to communicate intent more clearly, then we can use it where we want, but the footgun's safety is on by default.

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
JesusMcCloud: that is on me. keeps the binary compatibility surface smaller

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
n0900: we could do that but then we need to calculate `plainProtectedHeader` twice. Your call

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
n0900: In this case I think not actually because we never called the parameter serialzier?

JesusMcCloud: you are right, if no serialize call is made on the property we can scratch it

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
```diff
@@ -0,0 +1,133 @@
...
+ * This class does not support an unprotected header field!
+ *
+ * [JwsCompact] is intentionally *not* annotated with `@Serializable`: its JSON representation is only the compact
+ * JWS string, not a JSON object. Use [JwsCompactStringSerializer] explicitly when you want that string form inside
+ * a JSON document.
+ *
+ * For a standalone compact JWS string, use [toString] and [JwsCompact.invoke].
+ */
+data class JwsCompact(
+    val plainProtectedHeader: ByteArray,
```
JesusMcCloud, line 35: should this Ctor not be internal?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * A flattened JWS carries one payload and one signature. The protected header is stored as encoded bytes in
+ * [plainProtectedHeader]; the optional unprotected header is represented as [JwsHeader.Part]. The effective
+ * [jwsHeader] is reconstructed by merging both fragments with [JwsHeader.fromParts].
+ *
+ * Either header fragment may be partial. Only the combination of protected and unprotected parameters must
+ * constitute a valid [JwsHeader].
+ */
+@Serializable
+data class JwsFlattened(
```
JesusMcCloud: same here: should be internal, so we can only construct through signing from outside the module

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsGeneral.kt
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * General JSON JWS.
+ *
+ * A general JWS carries one payload and one or more [signatureElements]. Each [SignatureElement] contains the header
+ * fragments for one signature and exposes its merged effective [JwsHeader]. All signatures in a [JwsGeneral] share
+ * the same payload.
+ */
+@Serializable
+data class JwsGeneral(
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
```
JesusMcCloud, line 24: Pretty sure we want this here to also be internal and a companion object that pats inputs and ignature ops to signatureElements

JesusMcCloud (changes requested):
Minor API tweak requested

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
nodh: What if someone wants to write a serializer for this class, and then needs to call the constructor?

nodh:
@codex review please, you can do that!

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,125 @@
...
+         * The fragments may be partial, but their merged content must form a valid [JwsHeader].
+         */
+        suspend operator fun invoke(
+            protectedHeader: JwsHeader.Part?,
+            unprotectedHeader: JwsHeader.Part?,
+            payload: ByteArray,
+            signer: suspend (ByteArray) -> ByteArray
+        ): JwsFlattened {
+            val plainProtectedHeader = protectedHeader?.let { JwsProtectedHeaderSerializer.encodeToByteArray(it) }
+            return JwsFlattened(
```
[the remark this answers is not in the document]
nodh: That sounds like a serious issue

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsCompact.kt
JesusMcCloud: We have a serializer that handles canonical representation. I am against exposing it, because this here exists only to model this canonical representation

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JwsFlattened.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Either header fragment may be partial. Only the combination of protected and unprotected parameters must
+ * constitute a valid [JwsHeader].
+ */
+@Serializable
+data class JwsFlattened(
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
+    @SerialName(SerialNames.PROTECTED)
+    val plainProtectedHeader: ByteArray? = null,
+    @SerialName(SerialNames.HEADER)
+    val unprotectedHeader: JwsHeader.Part? = null,
```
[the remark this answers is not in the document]
n0900: This is true. Private Header Parameters will be supported when the use-case arises.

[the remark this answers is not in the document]
JesusMcCloud: we should document this fact

nodh (approved):

JesusMcCloud (approved):

PR #424, opened 2026-03-26
Fix/instant serialzer

n0900:
also the object exists twice 
indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/io/InstantLongSerializer.kt
indispensable-cosef/src/commonMain/kotlin/at/asitplus/signum/indispensable/cosef/InstantLongSerializer.kt

Unsure if there is a way to remedy this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## NEXT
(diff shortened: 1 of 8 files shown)


JesusMcCloud:
do those would-be duplicates behave the same and could be unified into `indispensable`?

JesusMcCloud (approved):
Add a changelog entry and we're good to go

PR #429, opened 2026-04-16
JWS: Add convenience protectedHeader parameter

n0900:
Useful since many standards require specific parameters to be part of the protected header specifically. Uniqueness of parameter already enforced via `.strictUnion` in eager `jwsHeader` parsing.

Cannot be parameter of `JWS` itself since `JwsGeneral` defines list of headers


JesusMcCloud (approved):
maybe add it as extension property instead?

PR #430, opened 2026-04-16
JwsHeader: ClientID

n0900:
For in the case of multisigned requests [OpenID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0-final.html#name-client_id) requires client_id to be part of the protected header


JesusMcCloud (approved):
lgtm

PR #431, opened 2026-04-20
Feature/unknown keys wrapper

n0900:
Adds generic wrapper to handle unknown keys which may be present in all json structures as 'private parameters' ([JWS for example](https://www.rfc-editor.org/rfc/rfc7515.html#section-4.3))
and an at-runtime get function.

Added `JwsHeader` and `JsonWebToken` as example usecases. Also looking for better names.
Might also be able to replace/move out domain specific parameters in the base classes which would allow null-checks on init where applicable. See `KeyAttestationJwt`

### DIFF
--- a/.github/config/test-strategy-matrix.json
+++ b/.github/config/test-strategy-matrix.json
@@ -22,6 +22,11 @@
       "testCommand": "./gradlew jsTest --scan"
     },
     {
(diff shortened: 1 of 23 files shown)


n0900:
Closed in favor of upcoming repository (Propigator)

PR #434, opened 2026-05-08
Feature/awesn2

JesusMcCloud:


JesusMcCloud:
@iaik-jheher no use to review before awesn1 PRs are done. I'll then push a new awesn1-release and adapt whatever's needed here based on awesn1-feedback

indispensable-cosef/src/jvmTest/kotlin/ConversionTests.kt
```diff
@@ -30,17 +29,19 @@ val ConversionTests by matrixSuite {
...
-                        val algorithm = x509.toCoseAlgorithm().getOrThrow()
+                        val algorithm = SignatureAlgorithm(x509).toCoseAlgorithm().getOrThrow()
                         algorithm shouldNotBe it
                         algorithm.toString().takeLast(3) shouldBe it.toString().takeLast(3)
                         algorithm.toString().take(2) shouldBe it.toString().take(2)
                     }
-                } else "is stable" { x509.toCoseAlgorithm() shouldSucceedWith it }
+                } else "is stable" {
+                    SignatureAlgorithm(x509).toCoseAlgorithm() shouldSucceedWith it
+                }
```
iaik-jheher:
```suggestion
                } else {
                    "is stable" {
                        SignatureAlgorithm(x509).toCoseAlgorithm() shouldSucceedWith it
                    }
                }
```

indispensable/src/androidMain/kotlin/at/asitplus/signum/indispensable/AndroidExtensions.kt
```diff
@@ -5,10 +5,10 @@ import java.security.Signature
...
 internal actual fun SignatureAlgorithm.RSA.getRSAPlatformSignatureInstance(provider: String?): Signature =
     when (this.padding) {
-        RSAPadding.PKCS1 ->
+        SignatureAlgorithm.RSA.Padding.PKCS1 ->
             sigGetInstance("${this.digest.jcaAlgorithmComponent}withRSA", provider)
 
-        RSAPadding.PSS ->
+        SignatureAlgorithm.RSA.Padding.PSS ->
             sigGetInstance("${this.digest.jcaAlgorithmComponent}withRSA/PSS", provider)
     }
```
iaik-jheher, line 13: is this not done via subclass switching on the parameters instance, now?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -1,40 +1,252 @@
...
+
+//for now, we just replicate the pattern, but since everything is sealed, we don't actually parse
+sealed interface SignatureAlgorithm : DataIntegrityAlgorithm, DerEncodable<X509AlgorithmIdentifier> {
+
+    //TODO: extensible
+    enum class Kind {
+        EC, RSA
+    }
+
+    val kind: Kind
```
iaik-jheher, line 39: why do we need this? why can't we just `is` check against the class?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -1,40 +1,252 @@
...
+
+    private data class EcdsaParams(val digest: Digest?, val curve: ECCurve?)
```
iaik-jheher: just stylistically I'd prefer if this was nested inside `ECDSA`, as `ECDSA.Params`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -47,29 +259,191 @@ sealed interface SignatureAlgorithm: DataIntegrityAlgorithm {
...
 
+            override fun decodeFromTlv(
+                serializer: KSerializer<X509AlgorithmIdentifier>,
+                src: Asn1Element,
+                der: Der
+            ) = RSA(der.decodeFromTlv(serializer, src))
+        }
+
+
+        sealed class Parameters<T : RsaParams> : DerEncodable<T> {
```
iaik-jheher: couldn't this be `sealed interface`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -47,29 +259,191 @@ sealed interface SignatureAlgorithm: DataIntegrityAlgorithm {
...
+
+            abstract val type: Padding
```
iaik-jheher, line 244: do we need this if we can just `is` check against the sealed interface?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -1,40 +1,252 @@
...
+        this === other -> true
+        this is RSA.Parameters.PssPadded && other is RSA.Parameters.PssPadded ->
+            digest == other.digest &&
+                    mgfAlgorithm == other.mgfAlgorithm &&
+                    saltLength == other.saltLength &&
+                    trailerField == other.trailerField
+
+        else -> this == other
+    }
+
```
iaik-jheher: isn't this exactly identical to just calling equals? the when branch with the type check just copies the contents of `PssPadded`'s `equals`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -1,40 +1,252 @@
...
+            if (!thisIsAsn1Backed && !otherIsAsn1Backed) {
+                return hasSamePropertiesAs(other)
+            }
+
+            if (asn1Representation == other.asn1Representation) return true
 
-        companion object : Enumeration<ECDSA> {
+            return runCatching {
+                hasSamePropertiesAs(other)
+            }.getOrDefault(false)
```
iaik-jheher: this seems to break `equals` transitivity, which is contractually required.

line 117 not being a `return` implies that two objects can be equal despite not having identical asn1 representation, as long as they parse identically. however, assume that object `A` is asn1 backed by a non-canonical encoding. now, `val B = SignatureAlgorithm(A.params)`. A and B compare equal. `val C = SignatureAlgorithm(B.asn1Representation)`. C is now asn1 backed by the canonical encoding. B and C compare equal. however, A and C are both asn1 backed with different encodings. A and C are therefore nonequal (line 109). this breaks transitivity.

you need to either compare all objects equal if they have the same content, or need to compare all objects equal only if they have identical asn1 representations. i suggest the former, and also suggest not masking exceptions encountered in the process.
```suggestion
            return hasSamePropertiesAs(other)
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Addons.kt
```diff
@@ -145,4 +59,95 @@ inline fun Asn1Primitive.decodeToBigIntegerOrNull(assertTag: Asn1Element.Tag = A
...
+    require(nBytes > 0) { "nBytes must be positive" }
+
+    val buf = ByteArray(nBytes)
+
+    do {
+        nextBytes(buf)
+    } while (!isDerMinimalIntegerEncoding(buf))
+
+    return Asn1Integer.fromTwosComplement(buf)
+}
```
iaik-jheher, line 79: doesn't `isDerMinimalIntegerEncoding` operate on asn.1 encoded bytes? but then why do we use `Asn1Integer.fromTwosComplement` which assumes a twos complement integer rather than the encoded form?

wouldn't it be far simpler to just generate `nBytes` random bytes, and parse a bigint from that? shouldn't the conversion strip off the prefix anyway?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Addons.kt
```diff
@@ -145,4 +59,95 @@ inline fun Asn1Primitive.decodeToBigIntegerOrNull(assertTag: Asn1Element.Tag = A
...
+        nextBytes(buf)
+
+        // Force positive sign bit.
+        buf[0] = (buf[0].toInt() and 0x7f).toByte()
+    } while (
+        buf.isAllZero || !isDerMinimalIntegerEncoding(buf)
+    )
+
+    return Asn1Integer.fromTwosComplement(buf) as Asn1Integer.Positive
+}
```
iaik-jheher: same question as above

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Addons.kt
```diff
@@ -145,4 +59,95 @@ inline fun Asn1Primitive.decodeToBigIntegerOrNull(assertTag: Asn1Element.Tag = A
...
+
+    do {
+        nextBytes(buf)
+
+        // Force negative sign bit.
+        buf[0] = (buf[0].toInt() or 0x80).toByte()
+    } while (!isDerMinimalIntegerEncoding(buf))
+
+    return Asn1Integer.fromTwosComplement(buf) as Asn1Integer.Negative
+}
```
iaik-jheher: same question as above

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Compat.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import at.asitplus.KmmResult
+import at.asitplus.KmmResult.Companion.wrap
+import at.asitplus.awesn1.Asn1Decodable
+import at.asitplus.awesn1.Asn1Element
+import at.asitplus.awesn1.Asn1Encodable
+import at.asitplus.awesn1.encoding.decodeFromDer
+import at.asitplus.awesn1.encoding.encodeToDer
+
+//TODO remove this file once
```
iaik-jheher: once what

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asymmetric/AsymmetricEncryptionAlgorithm.kt
```diff
@@ -5,7 +5,7 @@ import at.asitplus.signum.indispensable.Digest
 import at.asitplus.signum.Enumerable
 import at.asitplus.signum.Enumeration
 
-
+//TODO map to algorithmidentifier structure as per rfc8017
```
iaik-jheher, line 8: is this `TODO` still pending?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
```diff
@@ -89,19 +92,25 @@ object ByteArrayBase64UrlNoPaddingSerializer : TransformingSerializerTemplate<By
...
+object X509CertificateBase64Serializer : TransformingSerializerTemplate<Certificate, ByteArray>(
     parent = ByteArrayBase64Serializer,
-    encodeAs = X509Certificate::encodeToDer,
-    decodeAs = { X509Certificate.decodeFromDer(it) } // workaround iOS compilation bug KT-71498
+    encodeAs = ::encodeX509CertificateToDer,
+    decodeAs = ::decodeX509CertificateFromDer // workaround iOS compilation bug KT-71498
 )
 
+private fun encodeX509CertificateToDer(certificate: Certificate): ByteArray =
+    DER.encodeToTlv(Awesn1X509Certificate.serializer(), certificate.asn1Representation).derEncoded
```
iaik-jheher: this feels like a really awkward thing (always having to pass the serializer); isn't there a more convenient version of `encodeToTlv`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/Certificate.kt
```diff
@@ -0,0 +1,434 @@
...
+import at.asitplus.signum.internals.orLazy
+import io.matthewnelson.encoding.base64.Base64
+import io.matthewnelson.encoding.core.Decoder.Companion.decodeToByteArray
+import io.matthewnelson.encoding.core.Encoder.Companion.encodeToString
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Transient
+import kotlin.time.Clock
+import kotlin.time.Instant
+
+private data class TbsCertificateContent(
```
iaik-jheher: same as above, i would just semantically prefer if this was a nested class `TbsCertificate.ParsedContent` or similar

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CertificateExtension.kt
```diff
@@ -0,0 +1,155 @@
...
+            }
+            descriptors[descriptor.oid] = descriptor
+            return descriptor
+        }
+
+        fun descriptorFor(oid: ObjectIdentifier): Descriptor? = view()[oid]
+
+        private fun view(): Map<ObjectIdentifier, Descriptor> =
+            sealed.load() ?: descriptors.also { sealed.store(it) }
+    }
```
iaik-jheher, line 60: is this a necessary restriction? couldn't we just `if (sealed.load() != null) sealed.store(descriptors.copy())` or whatever and still maintain read access performance?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CertificationRequest.kt
```diff
@@ -0,0 +1,253 @@
...
+import at.asitplus.awesn1.crypto.pki.Pkcs10CertificationRequestInfo
+import at.asitplus.awesn1.serialization.DER
+import at.asitplus.awesn1.serialization.Der
+import at.asitplus.signum.indispensable.*
+import at.asitplus.signum.indispensable.pki.x500.X500Name
+import at.asitplus.signum.internals.orLazy
+import kotlinx.serialization.KSerializer
+import at.asitplus.awesn1.crypto.pki.X509CertificateExtension as Awesn1X509CertificateExtension
+
+private data class TbsCertificationRequestContent(
```
iaik-jheher: same as with others, i would prefer if this were nested

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CsrAttribute.kt
```diff
@@ -0,0 +1,91 @@
...
+        result = 31 * result + value.hashCode()
+        return result
+    }
+
+    override fun toString(): String = "X509CsrAttribute(oid=$oid, value=$value)"
+}
+
+internal fun CsrAttribute.requireX509(): CsrAttribute.X509Representable =
+    this as? CsrAttribute.X509Representable
+        ?: throw Asn1Exception("CSR attribute $oid has no X.509/DER representation")
```
iaik-jheher, line 91: i would prefer if this was just an `asn1Representation` extension property that does the type check and throws if mismatched

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/ExperimentalPkiApi.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Lives in the lean `indispensable` core (not `indispensable-pkix`) because core types such as
+ * [at.asitplus.signum.indispensable.pki.x500.GeneralName]'s constraint API are gated by it, and core
+ * cannot depend on `indispensable-pkix`.
+ */
+@RequiresOptIn(
+    message = "This API is part of the experimental certificate validation feature. " +
+            "It may not yet handle everything according to spec, could contain vulnerabilities, may change without notice, or eat your cat. " +
+            "Specify @OptIn(ExperimentalPkiApi::class)"
+)
+annotation class ExperimentalPkiApi(val message: String = "")
```
iaik-jheher, line 17: huh? is this part of this PR?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
```diff
@@ -0,0 +1,147 @@
...
+import at.asitplus.signum.indispensable.pki.ExperimentalPkiApi
+import at.asitplus.awesn1.Asn1Decodable
+import at.asitplus.awesn1.Asn1Element
+import at.asitplus.awesn1.Asn1Encodable
+import at.asitplus.awesn1.Asn1Exception
+import at.asitplus.awesn1.TagClass
+import kotlin.concurrent.atomics.AtomicReference
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+
+//TODO TODO TODO: Still needs to be ridden of ASN.1 specifics and made more generic! Still stays here for now a s plug for pkix extension module
```
iaik-jheher, line 13: is this `TODO` pending?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
```diff
@@ -0,0 +1,147 @@
...
+
+/** Builds an IMPLICIT context-specific primitive tag for a GeneralName CHOICE alternative. */
+fun contextTag(value: ULong) = Asn1Element.Tag(value, constructed = false, TagClass.CONTEXT_SPECIFIC)
```
iaik-jheher: this is publicly visible at file scope

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
```diff
@@ -0,0 +1,147 @@
...
+        fun register(): Descriptor = Registry.register(this)
+    }
+
+    /**
+     * Maps GeneralName CHOICE [NameType]s to their typed [Descriptor]s for the decode upgrade path.
+     *
+     * Registration is **startup-only**: descriptors must be registered (via [register], e.g. from
+     * `SignumPkix.install()`) **before the first (de)serialization**. The registry seals on its first
+     * lookup — after that it is immutable and reads are lock-free; later [register] calls throw.
+     */
```
iaik-jheher, line 109: same question regarding whether this is a necessary limitation

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
+import at.asitplus.catchingUnwrapped
+import at.asitplus.signum.indispensable.DerDecodable
+import at.asitplus.signum.indispensable.DerEncodable
+import at.asitplus.signum.internals.orLazy
+import kotlin.concurrent.atomics.AtomicReference
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+import kotlinx.serialization.KSerializer
+
+
+//TODO: this is a plug for pkix extension, needs further review separately
```
iaik-jheher, line 23: is this a pending `TODO`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
+            val atvs = rdnStr.splitRespectingEscapeAndQuotes('+').map { atvStr ->
+                val parts = atvStr.splitFirstUnescaped( '=')
+                if (parts.size != 2) throw IllegalArgumentException("Invalid RDN part: $atvStr")
+                AttributeTypeAndValue.fromString(parts[0], parts[1])
+                    ?: throw IllegalArgumentException("Unknown RDN part: $atvStr")
+            }
+            return RelativeDistinguishedName(atvs.toSet())
+        }
+
+        internal fun String.splitFirstUnescaped(delimiter: Char): List<String> {
```
iaik-jheher, line 101: couldn't this be `private`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
+            val regex = Regex("(?<!\\\\)${Regex.escape(delimiter.toString())}")
+            return this.split(regex, limit = 2)
+        }
 
-        constructor(str: Asn1String) : this(Asn1Primitive(str.tag, str.value.encodeToByteArray()))
+        internal fun String.splitRespectingEscapeAndQuotes(delimiter: Char): List<String> {
```
iaik-jheher, line 107: couldn't this be `private`

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
+
+
+
+abstract class BaseAttributeTypeAndValue(
+    override val oid: ObjectIdentifier,
+) : AttributeTypeAndValue {
+    override val displayName: String? get() = AttributeTypeAndValue.Registry.nameFor(oid)
+    override val isValid: Boolean? = null
+
+    override fun toString() = "AttributeTypeAndValue(oid=$oid)"
```
iaik-jheher: shouldn't this use `displayName` if available

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPrivateKey.kt
```diff
@@ -1,134 +1,186 @@
...
 import com.ionspin.kotlin.bignum.integer.Sign
-
-private object EB_STRINGS {
-    const val GENERIC_PRIVATE_KEY_PKCS8 = "PRIVATE KEY"
-    const val RSA_PRIVATE_KEY_PKCS1 = "RSA PRIVATE KEY"
-    const val EC_PRIVATE_KEY_SEC1 = "EC PRIVATE KEY"
-    const val ENCRYPTED_PRIVATE_KEY = "ENCRYPTED PRIVATE KEY"
+import kotlinx.serialization.KSerializer
+
+data class RsaPrivateKeyContent(
```
iaik-jheher: as before, would prefer nested class

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -162,17 +171,37 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
...
+            /** modulus */
+            n: Asn1Integer.Positive,
+
+            /** public exponent */
+            e: Asn1Integer.Positive,
+        ) : this(null, RsaPublicKeyContent(n, e))
+
+        constructor(asn1Representation: SubjectPublicKeyInfo) : this(asn1Representation, null)
+
+        // PEM disabled during awesn1 migration.
```
iaik-jheher: is this meant as a `TODO` for future work, or?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -257,15 +296,28 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
...
+        providedAsn1Representation: SubjectPublicKeyInfo?,
+        providedContent: EcPublicKeyContent?,
     ) : CryptoPublicKey(), KeyAgreementPublicValue.ECDH {
 
+        constructor(asn1Representation: SubjectPublicKeyInfo) : this(asn1Representation, null)
+
         override fun asCryptoPublicKey() = this
 
-        override val canonicalPEMBoundary: String = PEM_BOUNDARY
+        // PEM disabled during awesn1 migration.
```
iaik-jheher: same question

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
```diff
@@ -257,15 +296,28 @@ sealed class CryptoPublicKey : PemEncodable<Asn1Sequence>, Identifiable {
...
+            SubjectPublicKeyInfo.ec(curve.oid, iosEncoded)
+        }
+
+        val publicPoint: ECPoint.Normalized by providedContent?.publicPoint orLazy {
+            decodePublicPoint(asn1Representation)
+        }
+
+        val preferCompressedRepresentation: Boolean by providedContent?.preferCompressedRepresentation orLazy {
+            false
+        }
```
iaik-jheher:
```suggestion
        val preferCompressedRepresentation by providedContent?.preferCompressedRepresentation orLazy { false }
```

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
```diff
@@ -1,134 +1,100 @@
...
     ) : CryptoSignature {
 
-        init {
-            require(r.isPositive) { "r must be positive" }
-            require(s.isPositive) { "s must be positive" }
+        private val content: EcSignatureContent by providedContent orLazy {
+            providedAsn1Representation!!.decodeRS().let { (r, s) ->
+                EcSignatureContent(r.toBigInteger(), s.toBigInteger())
+            }
         }
```
iaik-jheher, line 60: this is a general feedback point on every class, not just here: didn't we agree that we'd put the conversion logic in a (faux) ctor on the parsed-contents class (so `EcSignatureContent` here) rather than putting the conversion line inline in the main class?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
```diff
@@ -137,138 +103,186 @@ sealed interface CryptoSignature : Asn1Encodable<Asn1Element> {
...
-            @Throws(Asn1Exception::class)
-            override fun doDecode(src: Asn1Element): RSA {
-                src as Asn1Primitive
-                return RSA(src)
+        operator fun invoke(kind: SignatureAlgorithm.Kind, asn1Representation: SignatureValue) =
+            when (kind) {
+                SignatureAlgorithm.Kind.EC -> EC(asn1Representation)
+                SignatureAlgorithm.Kind.RSA -> RSA(asn1Representation)
             }
-        }
```
iaik-jheher, line 259: i don't like this, it is not extension friendly. i'd prefer if this took a signaturealgorithm object instead, i think? the problem is we can't really make `SignatureAlgorithm.Kind` extensible

indispensable/src/jvmMain/kotlin/at/asitplus/signum/indispensable/JvmExtensions.kt
```diff
@@ -4,12 +4,11 @@ import java.security.Signature
...
 internal actual fun SignatureAlgorithm.RSA.getRSAPlatformSignatureInstance(provider: String?): Signature =
-    when (this.padding) {
-        RSAPadding.PKCS1 ->
+    when (this.parameters) {
+        is SignatureAlgorithm.RSA.Parameters.Pkcs1Padded ->
             sigGetInstance("${this.digest.jcaAlgorithmComponent}withRSA", provider)
 
-        RSAPadding.PSS -> sigGetInstance("RSASSA-PSS", provider).also {
+        is SignatureAlgorithm.RSA.Parameters.PssPadded -> sigGetInstance("RSASSA-PSS", provider).also {
             it.setParameter(this.digest.jcaPSSParams)
```
iaik-jheher: shouldn't this now be using parameters based on the actually parsed algorithm params? now that we parse those?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -47,29 +259,191 @@ sealed interface SignatureAlgorithm: DataIntegrityAlgorithm {
...
+                override val type: Padding get() = Padding.PSS
+                override val digest: Digest by providedParams?.digest orLazy {
+                    Digest.entries.first { it.oid == rsaSsaPssParams!!.effectiveHashAlgorithm.oid }
+                }
+
+                val mgfAlgorithm: MGF by providedParams?.mgfAlgorithm orLazy { MGF.entries.first { it.oid == rsaSsaPssParams!!.effectiveMaskGenAlgorithm.oid } }
+
+                val saltLength: UInt by providedParams?.saltLength orLazy {
+                    rsaSsaPssParams!!.effectiveSaltLength.let {
+                        require(it > 0)
```
iaik-jheher: shouldn't this be `>= 0`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
```diff
@@ -47,29 +259,191 @@ sealed interface SignatureAlgorithm: DataIntegrityAlgorithm {
...
+                    )
+
+                }
+
+                override val type: Padding get() = Padding.PSS
+                override val digest: Digest by providedParams?.digest orLazy {
+                    Digest.entries.first { it.oid == rsaSsaPssParams!!.effectiveHashAlgorithm.oid }
+                }
+
+                val mgfAlgorithm: MGF by providedParams?.mgfAlgorithm orLazy { MGF.entries.first { it.oid == rsaSsaPssParams!!.effectiveMaskGenAlgorithm.oid } }
```
iaik-jheher: doesn't this discard the digest in the `maskGenAlgorithm`'s parameters by only looking at `oid`?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
 
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
-        if (other == null || this::class != other::class) return false
+        if (other !is AttributeTypeAndValue) return false
+        return oid == other.oid
```
iaik-jheher, line 290: shouldn't this consider `value` too? if not, this seems to break symmetry with `BaseX509AttributeTypeAndValue` (`==` needs to be symmetric contractually); if A is a BaseAttributeTypeAndValue of another subclass type that does not override `equals`, and B is a BaseX509AttributeTypeAndValue with the same oid, then `A == B` will be true, but `B == A` will be false.

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/os/SigningProvider.kt
```diff
@@ -132,6 +132,9 @@ open class ECSignerConfiguration internal constructor(): DSL.Data() {
     internal var digestSpecified = false
 }
 open class RSASignerConfiguration internal constructor(): DSL.Data() {
+
+    //TODO: this is more convenient, but does not allow customization of PSS except for digest. would we allow to set parameters like in my previous commit,
+    //it would be exact, but nobody supports them anyways, so…
```
iaik-jheher: open `TODO`? though i will note that i don't think you can represent it to the platform anywhere

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "awesn1"]
+	path = awesn1
+	url = ../awesn1
```
iaik-jheher: this should not be here (i thought we do compounding via `../awesn1` composition where available?)

build.gradle.kts
```diff
@@ -46,6 +46,7 @@ subprojects {
 }
 
 allprojects {
+    repositories {mavenLocal()}
```
iaik-jheher: this feels like a leftover

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X500Name.kt
```diff
@@ -0,0 +1,137 @@
...
+                }
+                i++
+            }
+
+            val lastRdn = value.substring(start).trim()
+            if (lastRdn.isNotEmpty()) rdns.add(RelativeDistinguishedName.fromString(lastRdn))
+
+            return X500Name(rdns)
+        }
+    }
```
iaik-jheher, line 97: is this supposed to account for quotes? isn't `CN="Foo, Bar"` a valid DN?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
```diff
@@ -0,0 +1,147 @@
...
+         * the CHOICE tag (and decoding succeeds), else returning a generic [GeneralName] carrying the
+         * verbatim element. A typed-decode failure falls back to the generic form rather than throwing.
+         */
+        override fun doDecode(src: Asn1Element): GeneralName {
+            val type = NameType.fromTagValue(src.tag.tagValue)
+                ?: throw Asn1Exception("Unsupported GeneralName tag ${src.tag}")
+            return Registry.descriptorFor(type)
+                ?.let { catchingUnwrapped { it.fromTagged(src) }.getOrNull() }
+                ?: GeneralName(type, src)
+        }
```
iaik-jheher, line 141: this does not validate `constructed`/`CONTEXT_SPECIFIC`, which feels too permissive; for example, you could parse an integer as a name like this?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
iaik-jheher: also shouldn't some types be `constructed = true` (for sequence types such as `otherName`?)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X500Name.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    fun toRfc2253String(): String {
+        // RDN order is preserved; ATVs within a (multi-valued) RDN are sorted by attribute OID
+        // (DER encoding order, unsigned) for a canonical form.
+        return relativeDistinguishedNames.joinToString(",") { rdn ->
+            rdn.attrsAndValues
+                .sortedBy { it.oid }
+                .joinToString("+") { atv -> atv.toRfc2253String().trim() }
+        }
+    }
```
iaik-jheher, line 109: double check the order here (and in `fromString`). i believe the internal sequence is least specific first, while the string form should be most specific first?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CertificateExtension.kt
```diff
@@ -0,0 +1,155 @@
...
+                "CertificateExtension registry is sealed; register before the first (de)serialization."
+            }
+            descriptors[descriptor.oid] = descriptor
+            return descriptor
+        }
+
+        fun descriptorFor(oid: ObjectIdentifier): Descriptor? = view()[oid]
+
+        private fun view(): Map<ObjectIdentifier, Descriptor> =
+            sealed.load() ?: descriptors.also { sealed.store(it) }
```
iaik-jheher: `toMap` to make it immutable i think

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
```diff
@@ -1,138 +1,379 @@
...
+    override val asn1Representation: X500RelativeDistinguishedName by providedAsn1Representation orLazy {
+        X500RelativeDistinguishedName(attrsAndValues.map { it.requireX509().asn1Representation }.toSet())
+    }
+
+    val attrsAndValues: Set<AttributeTypeAndValue> by providedAttrsAndValues orLazy {
+        asn1Representation.attrsAndValues.map(AttributeTypeAndValue::fromAsn1Representation).toSet()
+    }
 
-    constructor(singleItem: AttributeTypeAndValue) : this(listOf(singleItem))
+    val isValid: Boolean by lazy { attrsAndValues.all { it.isValid != false } }
```
iaik-jheher, line 59: should this incorporate validate on the provideattrsandvalues equivalent (to catch multiple attributes with the same oid etc) since parsed names disable validation (i assume that part is intentional?)

iaik-jheher (changes requested):

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CsrAttribute.kt
```diff
@@ -0,0 +1,91 @@
...
+
+abstract class BaseCsrAttribute(
+    override val oid: ObjectIdentifier,
+) : CsrAttribute {
+    override fun toString(): String = "BaseCsrAttribute(oid=$oid)"
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is CsrAttribute) return false
+        return oid == other.oid
```
iaik-jheher, line 51: same issue as that other comment about asymmetric equals; if you have a subclass that does not override `equals` this will break the symmetry contract on `equals`

indispensable/src/androidMain/kotlin/at/asitplus/signum/indispensable/AndroidExtensions.kt
JesusMcCloud: this is just a shorthand that exposes the padding from the params, so it should amount to the same

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
JesusMcCloud: because this is half-baked. I was going to move this to the public and private keys and reference it from the sigalg, so we'd have an alignment between key types, and signature algos, but also check it for asymmetric encryption, and key agreement. WDYT?

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
JesusMcCloud: it is convenient to have, just like digest, and both are delegating to the actual value from the params, so I don't see this as bloat

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
JesusMcCloud: done. this removed even some more code

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
JesusMcCloud: removed

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asn1/Asn1Compat.kt
JesusMcCloud: removed. it already happened

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/asymmetric/AsymmetricEncryptionAlgorithm.kt
JesusMcCloud: Noted: https://github.com/a-sit-plus/signum/issues/456

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/io/Encoding.kt
JesusMcCloud: fixed

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CertificateExtension.kt
JesusMcCloud: i would very much like to enforce this. because registering custom serializers for open polymorphism, etc. in kotlinx.serialization is also only possible before the first access. if we don't align it here, people might happily register stuff here and once serialization kicks in, things may blow up, if it requires custom serializers as well.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CsrAttribute.kt
JesusMcCloud: you have me 90% convinced, as this would probably still make perfect sense for C509 too

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
JesusMcCloud: noted https://github.com/a-sit-plus/signum/issues/457

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
JesusMcCloud: has no business being here anyways. leftover from pkix extn.

removed

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: noted https://github.com/a-sit-plus/signum/issues/457

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: internal for tests (added a comment)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: internal for tests (added a comment)

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
JesusMcCloud: leftoever comment removed. it is already back

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
JesusMcCloud: leftoever comment removed. it is already back

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoPublicKey.kt
JesusMcCloud: applied manually

indispensable/src/jvmMain/kotlin/at/asitplus/signum/indispensable/JvmExtensions.kt
JesusMcCloud: it's just a delagate that amounts to the same result

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
JesusMcCloud: mgf foo fixed and tested

supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/os/SigningProvider.kt
JesusMcCloud: i removed the TODO prefix of the comment but kept the comment.

.gitmodules
JesusMcCloud: removed

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/CryptoSignature.kt
JesusMcCloud: general point. I think we need some marker property that aligins keys and algorithms, in an extensible way

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: https://github.com/a-sit-plus/signum/issues/458

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X500Name.kt
JesusMcCloud: https://github.com/a-sit-plus/signum/issues/458

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralName.kt
JesusMcCloud: https://github.com/a-sit-plus/signum/issues/458

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/X500Name.kt
JesusMcCloud: https://github.com/a-sit-plus/signum/issues/458

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/RelativeDistinguishedName.kt
JesusMcCloud: https://github.com/a-sit-plus/signum/issues/458

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CertificateExtension.kt
iaik-jheher: OK, maybe tweak the comment a bit, to me it reads as if this was a requirement for the lock-free reads.

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/CsrAttribute.kt
iaik-jheher: it's mainly a matter of: the `X509Representable` interface only exposes a single property, so `requireX509()` is just boilerplate. bonus points: the getter for the property should have a contract that smartcasts for you.

so, instead of removing this, make it public, and add a `val CsrAttribute.fooRepresentation get() = requireX509().fooRepresentation` except it has a contract if possible

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
iaik-jheher: It's generalizing away information about the padding; any variant of `PSS` turns into `Padding.PSS`. I prefer the type check.

indispensable-cosef/src/jvmTest/kotlin/ConversionTests.kt
iaik-jheher: this is still pending @JesusMcCloud

indispensable/src/commonMain/kotlin/at/asitplus/signum/indispensable/SignatureAlgorithm.kt
iaik-jheher: https://github.com/a-sit-plus/signum/pull/434#discussion_r3601201628 is a perfect illustration of how this leads to problems down the line

iaik-jheher (approved):
As I said async, approved conditional on the linked issues being fixed _imminently_ and _minimally_. No new structural rewrites, just those issues being fixed at the issue site.

PR #435, opened 2026-05-12
Feature/propigator josef jwt

n0900:
Initial idea on how to integrate Propigator into Signum and elsewhere.

Propigator-json allows us to resolve backed properties and slices from a general jsonObject. This is useful for instance for JWT where every usecase defines their own parameters and a general class which collects all becomes overloaded quickly.

I propose the following:
+ Keep semi up-to-date list of all IANA and other relevant claim names
+ Define seperate `*Payload` objects for each JWT use case (see `KeyAttestationPayload`)
+ Use `.validate()` for early aborts when applicable

Looking for feedback. If acceptable next steps would be doing the same for JWS-Header


indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ClientAttestationPayload.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+
+@Serializable(with = ClientAttestationPayload.Serializer::class)
+data class ClientAttestationPayload(
+    private val raw: JsonObject,
+    private val json: Json = joseCompliantSerializer,
+) : JsonObjectBacked(raw, JsonBackingCodec(json)), ObjectBackedValidated, JwtPayload {
+    val jwtClaims: JwtClaims by jsonSlice()
```
JesusMcCloud: Just that I understand correctly: This is built from a slice, but the object itself does not store a raw json object. How would I then add a new claim like so: `val JwtClaims.foobar get() = …`? Or is that irrelevant in this particular case, because Claims is essentially just a collection of known properties?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ConfirmationClaim.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlinx.serialization.Serializable
+
+/**
+ * Structure to declare posession of a particular proof-of-possesion key,
+ * to be included in [at.asitplus.signum.indispensable.josef.JwtClaims.confirmationClaim]
+ *
+ * See [RFC 7800](https://datatracker.ietf.org/doc/html/rfc7800)
+ */
+@Serializable
+data class ConfirmationClaim(
```
JesusMcCloud, line 14: But this one here might require extensibility or is it set in stone?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ClientAttestationPayload.kt
```diff
@@ -0,0 +1,39 @@
...
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonObject
+
+@Serializable(with = ClientAttestationPayload.Serializer::class)
+data class ClientAttestationPayload(
+    private val raw: JsonObject,
+    private val json: Json = joseCompliantSerializer,
+) : JsonObjectBacked(raw, JsonBackingCodec(json)), ObjectBackedValidated, JwtPayload {
```
JesusMcCloud: How would I construct it manually?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/KeyStorageStatus.kt
```diff
@@ -0,0 +1,24 @@
+package at.asitplus.signum.indispensable.josef.jwtpayload
+
+import at.asitplus.signum.indispensable.io.InstantLongSerializer
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonObject
+import kotlin.time.Instant
+
+@Serializable
+data class KeyStorageStatus(
```
JesusMcCloud, line 13: Here too. this feels like it should be extensible and this also jsonObject backed

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/WalletAttestationClaims.kt
```diff
@@ -0,0 +1,60 @@
...
+import at.asitplus.signum.indispensable.josef.JwtClaims.UnregisteredClaims
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * Wallet Instance Attestation (WIA) as defined by
+ * [EUDI TS3](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md)
+ */
+@Serializable
+data class WalletAttestationClaims(
```
JesusMcCloud, line 14: Especially a moving target, it it not? Is thsi class not precisely the kinf ot type we'd tant to bolt-on additional orpoerties through extensions?

JesusMcCloud (commented):
left some generic comments. I think we should discuss in a call offline

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ClientAttestationPayload.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    constructor(
+        jwtBase: JwtBaseClaims,
+        confirmationClaim: ConfirmationClaim,
+        misc: JsonObject,
```
JesusMcCloud: i am torn: jwtBase and confirmationClaim are flattened, so it makes sense to have misc as an object. then again, vararg Pair<String, JsonElement> would be closer to the wire model if those are the just plugged into the root node

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ClientAttestationPayload.kt
n0900: Added secondary constructors

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/ClientAttestationPayload.kt
n0900: no it stores a raw jsonObject, the slice is just a predefined set of jsonBacked parameters. If you need an additional parameter you would extend `ClientAttestationPayload` directly and not the slice

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/jwtpayload/KeyStorageStatus.kt
n0900: The payload itself is jsonBacked and can be extended, this is just a convenience class to define slices

JesusMcCloud:
still in progress, right?

PR #439, opened 2026-05-21
Feature/jws typed serializer

n0900:
The desire-path strikes again.

Thin serializer template that goes through JwsTyped to the over-the-wire format JWS for correct serialization. Might actually be more performant than normal usage because types and serializers need to be known at compile time? (P no longer reified)

Edit: I am very sorry for the generics re-ordering


JesusMcCloud (commented):
why the reverse of generics?

n0900:
> why the reverse of generics?

Because its this way around for all other instances. Makes for bad user experience. My bad i missed that orginially

PR #444, opened 2026-06-01
Implement JWE shapes as for JWS

n0900:
Refactors JWE in the same way as JWS also implementing the three officially supported over-the-wire formats.
`JWE` is a generalization of `JweEncrypted` whereas `JweTyped` (not implemented yet) is the generailzation of `JweDecrypted`. This is of renewed interest for us because [JOSE HPKE](https://datatracker.ietf.org/doc/html/draft-ietf-jose-hpke-encrypt-16#name-flattened-jwe-json-serializ) also allows for multiple wire-formats.

Since in many applications (most notably JsonWebToken [RFC 7519](https://www.rfc-editor.org/info/rfc7519/)) support JWS as well as JWE for their secured data transfers this also introduces `JsonSecured` as a super-type and a delegating serializer.
 
TODO:
- [x] Implement `JWE`
- [x] Implement `JweTyped`

~~Builds on #435~~ Now independent


indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweCompact.kt
```diff
@@ -0,0 +1,143 @@
...
+         */
+        operator fun invoke(base64UrlString: String): JweCompact {
+            require(!base64UrlString.contains("=")) { "Trailing = are not supported. See RFC 7516" }
+            val partCount = base64UrlString.count { it == '.' } + 1
+            if (partCount != 5) {
+                throw SerializationException(
+                    "Invalid JWE compact serialization: expected 5 parts, got $partCount"
+                )
+            }
+            val parts = base64UrlString.split('.', limit = 5)
```
iaik-jheher: why not just `.split` and then check the length?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweCompact.kt
```diff
@@ -0,0 +1,143 @@
...
+        result = 31 * result + (authenticationTag?.contentHashCode() ?: 0)
+        return result
+    }
+
+    companion object {
+        /**
+         * Build a [JweCompact] received as compact serialization string.
+         */
+        operator fun invoke(base64UrlString: String): JweCompact {
+            require(!base64UrlString.contains("=")) { "Trailing = are not supported. See RFC 7516" }
```
iaik-jheher, line 72: this feels like a (de)serializer property (allowing/disallowing padding) rather than an external string check here

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
```diff
@@ -0,0 +1,231 @@
...
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.Transient
+import kotlinx.serialization.json.JsonObject
+
+/**
+ * Flattened JSON JWE serialization.
+ *
+ * A flattened JWE carries one encrypted content value and one recipient. Unprotected header parameters can be split
+ * between the shared [sharedUnprotectedHeader] and per-recipient [recipientUnprotectedHeader] locations.
```
iaik-jheher, line 14: what does the comment regarding header params mean? since this form only has one recipient, aren't those locations interchangable?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
```diff
@@ -0,0 +1,231 @@
...
+ */
+@ConsistentCopyVisibility
+@Serializable
+data class JweFlattened internal constructor(
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
+    @SerialName(SerialNames.PROTECTED)
```
iaik-jheher, line 20: do we usually place serial names in constants? this feels like an unhelpful layer of abstraction to me, they only belong in one place (here) and having to hover/look at the constant makes it unclear what's happening

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
```diff
@@ -0,0 +1,231 @@
...
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
+    @SerialName(SerialNames.CIPHERTEXT)
+    val ciphertext: ByteArray,
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
+    @SerialName(SerialNames.AUTHENTICATION_TAG)
+    val authenticationTag: ByteArray? = null,
+) : JWE() {
+
+    init {
+        JweProtectedHeaderSerializer.requireAbsentIfEmpty(plainProtectedHeader)
```
iaik-jheher: does this form do something different from the non-class form used in the following lines?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
```diff
@@ -0,0 +1,231 @@
...
+            additionalAuthenticatedData: ByteArray? = null,
+            initializationVector: ByteArray? = null,
+            ciphertext: ByteArray,
+            authenticationTag: ByteArray? = null,
+        ): JweFlattened {
+            val plainProtectedHeader = JweProtectedHeaderSerializer.encodeToByteArrayOrNull(protectedHeader)
+            return JweFlattened(
+                plainProtectedHeader = plainProtectedHeader,
+                sharedUnprotectedHeader = sharedUnprotectedHeader?.takeUnless { it.toJsonObject().isEmpty() },
+                recipientUnprotectedHeader = recipientUnprotectedHeader?.takeUnless { it.toJsonObject().isEmpty() },
```
iaik-jheher: is that not just `.takeUnlessEmpty()` again?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
```diff
@@ -0,0 +1,231 @@
...
+    recipientUnprotectedHeader: JweHeader.Part?,
+): JweHeader.Part? {
+    val merged = sharedUnprotectedHeader.normalizedJsonObject()
+        .strictUnion(recipientUnprotectedHeader.normalizedJsonObject())
+    return merged.takeIf { it.isNotEmpty() }
+        ?.let { joseCompliantSerializer.decodeFromJsonElement(JweHeader.Part.serializer(), it) }
+}
+
+internal fun JweHeader.Part?.normalized(): JweHeader.Part? =
+    this?.takeUnless { it.toJsonObject().isEmpty() }
```
iaik-jheher: didn't we call this `.takeUnlessEmpty` earlier

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * General JSON JWE serialization.
+ *
+ * A general JWE carries one encrypted content value and one or more [RecipientElement]s.
+ */
+@ConsistentCopyVisibility
+@Serializable
+data class JweGeneral internal constructor(
+    @Serializable(ByteArrayBase64UrlNoPaddingSerializer::class)
+    @SerialName(SerialNames.PROTECTED)
```
iaik-jheher, line 18: same comment as before regarding serial names

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Transient
+    val additionalAuthenticatedDataInput: ByteArray =
+        getAdditionalAuthenticatedData(plainProtectedHeader, additionalAuthenticatedData)
+
+    /**
+     * Returns a new [JweGeneral] with one additional recipient over the same encrypted content.
+     */
+    fun appendRecipient(jweFlattened: JweFlattened): JweGeneral {
```
iaik-jheher, line 63: wouldn't it make more sense to just be able to append a `RecipientElement` instead?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
```diff
@@ -0,0 +1,163 @@
...
+                encryptedKey = jweFlattened.encryptedKey,
+            )
+        )
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
+
+        other as JweGeneral
```
iaik-jheher, line 87: `if (other !is JweGeneral)`?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
```diff
@@ -0,0 +1,163 @@
...
+            recipientUnprotectedHeaders.map { recipientUnprotectedHeader ->
+                JweFlattened(
+                    protectedHeader = protectedHeader,
+                    payload = payload,
+                    sharedUnprotectedHeader = sharedUnprotectedHeader,
+                    recipientUnprotectedHeader = recipientUnprotectedHeader,
+                    additionalAuthenticatedData = additionalAuthenticatedData,
+                    encryptor = encryptor,
+                )
+            }.toJweGeneral()
```
iaik-jheher: does this make sense? to build a bunch of `JweFlattened`s and then merge them? not saying it doesn't, just unsure. PS: what if the encryption isn't deterministic? wouldn't that make the merging fail?

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweHeader.kt
```diff
@@ -328,11 +447,98 @@ data class JweHeader(
...
+            protectedHeader: JsonObject? = null,
+            sharedUnprotectedHeader: JsonObject? = null,
+            recipientUnprotectedHeader: JsonObject? = null,
+        ): JweHeader = joseCompliantSerializer.decodeFromJsonElement<JweHeader>(
+            protectedHeader
+                .strictUnion(sharedUnprotectedHeader)
+                .strictUnion(recipientUnprotectedHeader)
+        ).also {
+            requireNotNull(it.algorithm) { "JWE JOSE Header must contain 'alg'" }
+            requireNotNull(it.encryption) { "JWE JOSE Header must contain 'enc'" }
```
iaik-jheher: if we are doing strict RFC for now this is correct, though if we implement the HPKE draft i will note that this must be present if and only if the algorithm is not an integrated encryption algorithm (such as non-KE HPKE)

iaik-jheher (commented):
Additionally, I understand that the split is kind of strange to deal with (since you don't have `supreme` here) but are we sure this setup with the `encryptor`/`decryptor` lambdas makes sense? given that those lambdas need to make decisions based on the contents of the headers and etc, it feels like a bit of magic-black-box handwaving to me that should maybe be implemented in a signum module. @JesusMcCloud this ties into the discussion we just hadd ofline

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweCompact.kt
n0900: I agree but here its though because this is a public constructor. To have the constructor makes sense because there are many situations where a json element is sent over the wire but it may be a plain `JsonElement`, a `JWS` or a `JWE` so you have to receive it as string and invoke the constructor manually

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
n0900: yes it doesnt make sense, but thats the [spec](https://datatracker.ietf.org/doc/html/rfc7516#section-7.2.2)

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
n0900: This was started some time ago in VCK and I'm just continuing this pattern here for consistency. Can be removed.

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
n0900: the plain protected header is a base64 string/bytearray so in theory it could encode the empty json object whereas `sharedUnprotectedHeader` has already been deserialized and can be verified to be non-empty but this should probably be solved differently

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
n0900: As above, theoretically `recipientUnprotectedHeader` could be the empty `JweHeader.Part` but this should also be solved differently

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweFlattened.kt
n0900: ~~`.takeUnlessEmpty` targets `ByteArray` but yes, should be streamlined~~
yes

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
n0900: Autogenerated by intellij but yes we can do that. Should be captured by `  if (other == null || this::class != other::class) return false` though

indispensable-josef/src/commonMain/kotlin/at/asitplus/signum/indispensable/josef/JweGeneral.kt
n0900: I think my problem was that just appending a `RecipientElement` doesnt perform any checks. 
Right now for the library user to get a `JweFlattened` it has to be either a) explicitly produced (assuming the `encryptor` is a valid implementation) correctly or b) received over the wire.

Otherwise we need to either make sure that the correct CEK is used for the payload and emitted as `encrypted_key` checking this inside the function or give no guarantees at all.

n0900:
`encryptor`/`decryptor` are a pain point for me as well but we need to expose some constructor to the outside and the primary constructor is difficult to use. The same issue arises for `JwsCompact` etc

PR #453, opened 2026-06-23
Feature/pkix

JesusMcCloud:
merge @StjepanovicSrdjan's cert path validation on top of awesn1-backed signum.

This is only the merge, not yet a rewrite based on kotlinx.serializazion and/or discrete modules. Registry allows for externalising nicely


JesusMcCloud:
source was https://github.com/a-sit-plus/signum/pull/413

StjepanovicSrdjan (commented):
lgtm

indispensable-pkix/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/extn/AuthorityKeyIdentifier.kt
```diff
@@ -1,44 +1,79 @@
...
+ *   keyIdentifier             [0] KeyIdentifier             OPTIONAL,   -- OCTET STRING
+ *   authorityCertIssuer       [1] GeneralNames              OPTIONAL,   -- SEQUENCE OF GeneralName
+ *   authorityCertSerialNumber [2] CertificateSerialNumber   OPTIONAL }  -- INTEGER
+ * ```
+ * All members are IMPLICIT context-tagged. `keyIdentifier` is a `ByteArray` (awesn1 encodes it as the
+ * OCTET STRING); [1]/[2] infer their tags from the `List` / `Asn1Integer` descriptors. Shared by decode
+ * and the programmatic constructor.
+ */
+@Serializable
+private class AuthorityKeyIdentifierBody(
```
wkornewald, line 76: Not sure if this is solved differently in the latest KxS but we always hat problems if Serializable isn't at least internal because the companion object gets extended by the compiler plugin and thus inline reified KxS APIs don't work. Above you're passing the serializer explicitly, so maybe it's fine, but internal is less error-prone than private.

indispensable-pkix/src/commonMain/kotlin/at/asitplus/signum/indispensable/pki/GeneralSubtree.kt
```diff
@@ -0,0 +1,277 @@
...
+    }
+}
+
+/**
+ * A `GeneralSubtrees ::= SEQUENCE SIZE (1..MAX) OF GeneralSubtree` plus the RFC 5280 name-constraint
+ * merge/minimize logic. (De)serialization is handled at the containing [NameConstraints] as a
+ * `List<GeneralSubtree>` under its `[0]`/`[1]` field, so this type carries no ASN.1 (de)coding of its own.
+ */
+data class GeneralSubtrees(
+    var trees: MutableList<GeneralSubtree>
```
wkornewald, line 65: Why MutableList?

pkix-supreme/src/commonMain/kotlin/at/asitplus/signum/supreme/validate/TimeValidityValidator.kt
```diff
@@ -0,0 +1,54 @@
...
+                    dateOfIssuance = childCert.tbsCertificate.validFrom,
+                    issuer = currCert,
+                    processingChain.size - 1 - currentCertIndex)
+            }
+        }
+
+        return emptyMap()
+    }
+
+    private fun wasCertificateIssuedWithinIssuerValidityPeriod(
```
wkornewald, line 42: In practice, we've hit problems with the check for `issuer.notBefore < child.notBefore` e.g. for cross-signing certificates which can be created after the child was issued. Maybe it would be better to simply compute the validity period by taking the issuer chain into account and narrow the interval?

JesusMcCloud:
contains a lot of hardcoded `DER` references. need to envistigate

PR #462, opened 2026-07-23
Improve docs in JwsTyped

n0900:
Small Doc change


JesusMcCloud (approved):
if it really is that important…

PR #465, opened 2026-07-27
conditionally include awesn1 composite build

JesusMcCloud:
tested locally by manually editing the version number of awesn1, cloned into signum.

merge at your discretion


iaik-jheher:
shouldn't we be using `../awesn1`? maybe i am misunderstanding how we expect this compositing to work, but I thought we used side-by-side directories

iaik-jheher:
see e.g. how vc-k does it:
image

iaik-jheher:
changed after offline discussion

iaik-jheher (approved):

PR #468, opened 2026-08-07
Extend JsonWebToken

nodh:
Kindly requesting inclusion in a point release onto 3.24.0

Needed to support latest draft of [OAuth 2.0 Attestation-Based Client Authentication](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-10)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## NEXT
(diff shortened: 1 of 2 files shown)


JesusMcCloud:
@nodh it's live (and will propagate on maven central within the hour): https://github.com/a-sit-plus/signum/releases/tag/3.25.0

PR #473, opened 2026-08-19
Expose iOS authentication timeout

### LINKED ISSUE #472: [Supreme/iOS] Expose authentication timeout for restored signers
## Problem

`IosSigner` exposes whether authentication is required and whether every use requires it, but not the positive authentication-reuse timeout restored from Signum key metadata. Downstream callers therefore cannot verify that a restored signer retained its configured reuse window.

Downstream context: [walt-id/waltid-identity#2124](https://github.com/walt-id/waltid-identity/pull/2124) explicitly models the current timeout-readback limitation. That temporary workaround can be removed once this enhancement is available upstream.

## Proposed API

Expose the effective authentication timeout as a read-only `Duration?`:

- `null`: authentication is not required
- `Duration.ZERO`: authenticate for every use
- positive duration: Signum may reuse authentication for that interval

The value should come from Signum's stored key metadata. It does not represent independent iOS Keychain access-control readback and should not change enforcement or caching behavior.

## Acceptance criteria

- The timeout is available on created and restored iOS signers.
- Existing authentication booleans remain consistent with it.
- Metadata round-trip tests cover no authentication, every-use authentication, and a positive timeout.

szijpeter:
## Summary

Expose `IosSigner.authenticationTimeout` so callers can distinguish keys that require no authentication, authentication for every use, or authentication with a positive reuse window—even after signer restoration.

The value comes from Signum's stored key metadata. It does not change authentication enforcement or caching and does not claim independent iOS Keychain access-control readback. The existing authentication booleans now derive from the same property.

Closes #472.

## Tests

- `:supreme:iosSimulatorArm64Test`
- `:supreme:compileKotlinIosArm64`

### DIFF
--- a/supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
+++ b/supreme/src/iosMain/kotlin/at/asitplus/signum/supreme/os/IosKeychainProvider.kt
@@ -130,8 +130,24 @@ sealed class IosSigner(final override val alias: String,
     override suspend fun exportPrivateKey(): KmmResult<Nothing> = KmmResult.failure(IllegalStateException("Non-Exportable key"))
 
     override val mayRequireUserUnlock get() = needsAuthentication
-    val needsAuthentication get() = metadata.needsUnlock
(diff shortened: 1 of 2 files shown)


JesusMcCloud:
thank you! Will be added on top of #475

PR #474, opened 2026-08-20
Implement cancellation handling for signing operations in Android, iOS key and JVM providers

AlonsoFloo:
## Summary

Adds coroutine cancellation support to signing operations across Android and iOS.
* Propagates coroutine cancellation instead of wrapping it as a signing error.
* Cancels active biometric authentication when the coroutine is cancelled.
* Adds tests covering cancellation and timeout scenarios.

(This PR has been assisted by AI)

## Reasoning

The old implementation did not allow an application to cancel a `.sign()` call when it involved **biometric authentication**.

Depending on the use case and the device, it can be useful to cancel the suspended function for example, when a timeout is reached or the user navigates away from the current screen.

## Example

The signing operation can now be cancelled using the standard coroutine timeout mechanism:

```kotlin
val result = withTimeout(5.seconds) {
    signer.sign(data)
}
```

It can also be wrapped in a `Flow` and combined with the existing `timeout` operator:

```kotlin
flow {
    emit(signer.sign(data))
}
    .timeout(5.seconds)
    .collect { result ->
        // Handle signature result
    }
```


JesusMcCloud:
Thank you! We will carry this over onto #475

